PR #70, opened 2015-01-27
Add group resource

a2ikm:
This example below will create a group named "foo" with gid 501.

```rb
group "foo" do
  gid 501
end
```

group in specinfra is introduced as of 2.12.

https://github.com/serverspec/specinfra/pull/303


ryotarai:

ryotarai:
This is released in v1.1.3.

PR #71, opened 2015-01-29
Make `node` handle NameError from host_inventory

nownabe:
I found two errors of `node`.

First one is as follows.
It is found with `node["undefined"]`:

```
/Users/nownabe/itamae-test/vendor/bundle/ruby/2.2.0/gems/specinfra-2.12.1/lib/specinfra/host_inventory.rb:22:in `const_get': uninitialized constant Undefined (NameError)
	from /Users/nownabe/itamae-test/vendor/bundle/ruby/2.2.0/gems/specinfra-2.12.1/lib/specinfra/host_inventory.rb:22:in `[]'
	from /Users/nownabe/itamae-test/vendor/bundle/ruby/2.2.0/gems/itamae-1.1.3/lib/itamae/node.rb:19:in `[]'
	from /Users/nownabe/itamae-test/recipe.rb:1:in `initialize'
```

Second one is as follows.
It is found with `node["un-defined"]`:

```
/Users/nownabe/itamae-test/vendor/bundle/ruby/2.2.0/gems/specinfra-2.12.1/lib/specinfra/host_inventory.rb:22:in `const_get': wrong constant name Un-defined (NameError)
	from /Users/nownabe/itamae-test/vendor/bundle/ruby/2.2.0/gems/specinfra-2.12.1/lib/specinfra/host_inventory.rb:22:in `[]'
	from /Users/nownabe/itamae-test/vendor/bundle/ruby/2.2.0/gems/itamae-1.1.3/lib/itamae/node.rb:19:in `[]'
	from /Users/nownabe/itamae-test/recipe.rb:1:in `initialize'
```

### DIFF
--- a/lib/itamae/node.rb
+++ b/lib/itamae/node.rb
@@ -17,7 +17,7 @@ module Itamae
       if val.nil?
         begin
           val = host_inventory[key]
-        rescue NotImplementedError
+        rescue NotImplementedError, NameError
           val = nil
         end
       end
--- a/spec/integration/default_spec.rb
+++ b/spec/integration/default_spec.rb
@@ -34,6 +34,7 @@ describe file('/tmp/template') do
   its(:content) { should match(/Hello/) }
   its(:content) { should match(/Good bye/) }
   its(:content) { should match(/^total memory: \d+kB$/) }
+  its(:content) { should match(/^uninitialized node key: $/) }
 end
 
 describe file('/tmp/file') do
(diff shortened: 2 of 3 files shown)


ryotarai:
I also think `#[]` with unset key should return `nil` and I'll merge this.

@mizzy I think Specinfra's `HostInventry#[]` should return `nil` instead of raising `NameError` or `NotImplementedError` when the key is not found, like `Hash#[]`. How about you?

ryotarai:
This is released in v1.1.4.

mizzy:
> @mizzy I think Specinfra's HostInventry#[] should return nil instead of raising NameError or NotImplementedError when the key is not found, like Hash#[]. How about you?

You're right. I will fix it later. Thanks!

PR #90, opened 2015-03-02
Add source option to gem_package resource

eagletmt:
Let users install gems from another gem server.
http://guides.rubygems.org/command-reference/#gem-install


ryotarai:
Thank you!

PR #91, opened 2015-03-06
remove reqirement of --host option for ssh command

muratayusuke:
This pull request includes a proposal and a bugfix.

### Proposal

I don't think we need `--host` option when `--vagrant` is provided. Just `itamae ssh --vagrant recipi.rb` is enough. So I remove `required: true` from ssh command and add validation instead.

### Bugfix

`vagrant ssh-config #{opts[:host]}` is invalid command on latest version of vagrant (1.7.2). Available commands are `vagrant ssh-config` and `vagrant ssh-config --host <hostname>`.

    $ vagrant ssh-config
    Host default
      HostName xxxx
      User xxxx
      Port xxxx
      UserKnownHostsFile /dev/null
      StrictHostKeyChecking no
      PasswordAuthentication no
      IdentityFile /path/to/keyfile
      IdentitiesOnly yes
      LogLevel FATAL

    $ vagrant ssh-config --host somehost
    Host somehost
      HostName xxxx
      User xxxx
      Port xxxx
      UserKnownHostsFile /dev/null
      StrictHostKeyChecking no
      PasswordAuthentication no
      IdentityFile /path/to/keyfile
      IdentitiesOnly yes
      LogLevel FATAL

In this pull request, I want to ignore `--host` option, so just use `vagrant ssh-config` and settings for `default` host.

I hope I can get any feedback!

### DIFF
--- a/lib/itamae/runner.rb
+++ b/lib/itamae/runner.rb
@@ -69,8 +69,9 @@ module Itamae
 
           if options[:vagrant]
             config = Tempfile.new('', Dir.tmpdir)
-            `vagrant ssh-config > #{config.path}`
-            opts.merge!(Net::SSH::Config.for('default', [config.path]))
+            hostname = opts[:host] || 'default'
+            `vagrant ssh-config #{hostname} > #{config.path}`
+            opts.merge!(Net::SSH::Config.for(hostname, [config.path]))
             opts[:host] = opts.delete(:host_name)
           end
 
--- a/lib/itamae/runner.rb
+++ b/lib/itamae/runner.rb
@@ -69,8 +69,8 @@ module Itamae
 
           if options[:vagrant]
             config = Tempfile.new('', Dir.tmpdir)
-            `vagrant ssh-config #{opts[:host]} > #{config.path}`
-            opts.merge!(Net::SSH::Config.for(opts[:host], [config.path]))
+            `vagrant ssh-config > #{config.path}`
+            opts.merge!(Net::SSH::Config.for('default', [config.path]))
             opts[:host] = opts.delete(:host_name)
           end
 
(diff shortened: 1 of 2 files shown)


lib/itamae/runner.rb
```diff
@@ -69,8 +69,8 @@ def set_backend_from_options(type, options)
 
           if options[:vagrant]
             config = Tempfile.new('', Dir.tmpdir)
-            `vagrant ssh-config #{opts[:host]} > #{config.path}`
-            opts.merge!(Net::SSH::Config.for(opts[:host], [config.path]))
+            `vagrant ssh-config > #{config.path}`
+            opts.merge!(Net::SSH::Config.for('default', [config.path]))
```
ryotarai: This change can't work with multi vagrant VMs.

ryotarai:
> vagrant ssh-config #{opts[:host]} is invalid command on latest version of vagrant (1.7.2). Available commands are vagrant ssh-config and vagrant ssh-config --host <hostname>.

`vagrant ssh-config --host name` is different from `vagrant ssh-config name`. `vagrant ssh-config name` is for multiple VMs.

muratayusuke:
@ryotarai Thank you for your review. I didn't know about multiple VMs setting and usage of `vagrant ssh-config name`. Now I can understand them (I saw this page: http://docs.vagrantup.com/v2/vagrantfile/tips.html), so I changed my code to restore original function and just accept `itamae ssh --vagrant recipi.rb` without hostname.

 I've never use multiple VMs, so I couldn't understand what hostname should I pass to the command when I use `itamae ssh` first time. I want to run this command without hostname when I have only `default` VM. How do you think about it?

ryotarai:
This change is acceptable. Thank you for your contribution!

PR #92, opened 2015-03-06
Add remove action to package resource

eagletmt:
It requires recent specinfra.
https://github.com/serverspec/specinfra/pull/335


ryotarai:
Thank you!

PR #96, opened 2015-03-14
[WIP] Add ssh_import_github attribute to user resource

nownabe:
Hi. My proposal is importing ssh keys from github.

In this pull request, user resource can get ssh public keys from github with `ssh_import_github` attribute.

I have not implemented showing difference of `authorized_keys` and tests yet.
So, I think that user resouce should use file resouce to create ssh keys if possible.

If this proposal is acceptable, please teach me best way to create ssh keys from github.

### DIFF
--- a/lib/itamae/resource/user.rb
+++ b/lib/itamae/resource/user.rb
@@ -1,4 +1,5 @@
 require 'itamae'
+require 'open-uri'
 
 module Itamae
   module Resource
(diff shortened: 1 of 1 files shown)


ryotarai:
Thank you for your PR.

I think this feature is not general one and it shouldn't be included in Itamae itself.
How about creating a gem as a plugin?

nownabe:
Thank you for your review!

Creating gem sounds good!
I'll try to implement this feature as plugin gem.

I proposed this feature reffering [CoreOS cloud-config](https://coreos.com/docs/cluster-management/setup/cloudinit-cloud-config/#users) just for your information.

PR #97, opened 2015-03-16
Ignore CommandExecutionError in dry-run mode

eagletmt:
When Ruby will be installed by another recipe but the recipe isn't
applied yet, `gem_package` resource with dry-run mode raises an error
because `gem list -l` command fails. CommandExecutionError should be
ignored during getting installed gems.


ryotarai:
Good catch :+1:  Thank you.

PR #103, opened 2015-03-18
Make node accessible in define block.

ryotarai:


ryotarai:
All tests are passed locally.

ryotarai:
@sorah fixed

PR #105, opened 2015-03-20
Make logging less verbose by default

eagletmt:
The default itamae output looks too verbose to me. My proposal is making
resource's output less verbose; it now displays output only if the
action is taken place.

https://gyazo.wanko.cc/3eeb9a6e88c2ecafdc7f7db0f5a98ccc

### DIFF
--- a/lib/itamae/resource/base.rb
+++ b/lib/itamae/resource/base.rb
@@ -116,26 +116,22 @@ module Itamae
       end
 
       def run(specific_action = nil, options = {})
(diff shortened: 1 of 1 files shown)


ryotarai:
Thank you for your PR. 
This PR is acceptable but I think [making indentation](https://github.com/itamae-kitchen/itamae/pull/105/files#diff-6fb90b872eacc75053b7a7db2924966fL121) is better. I'll try implement it after merging this.

PR #111, opened 2015-03-23
Fix for Vagrant

mizzy:
I've found that v1.2.2 does not work with Vagrant. So I've fixed it.

Please check it.


ryotarai:
Thank you!

PR #115, opened 2015-03-30
Add support to use ssh-config host

maruware:
This Pull Request will support to use host written in ssh config when executed with ssh option.
```
bundle exec itamae ssh -h host_in_ssh_config  recipe.rb
```

### DIFF
--- a/lib/itamae/backend.rb
+++ b/lib/itamae/backend.rb
@@ -1,6 +1,7 @@
 require 'specinfra/core'
 require 'singleton'
 require 'io/console'
+require 'net/ssh'
 
 Specinfra::Configuration.error_on_missing_backend_type = true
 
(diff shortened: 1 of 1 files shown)


ryotarai:
Good catch! I think the following is better.
What do you think?

```diff
-        opts[:user] = @options[:user] || Etc.getlogin
+
+        # from ssh-config
+        opts.merge!(Net::SSH::Config.for(@options[:host]))
+        opts[:user] = @options[:user] || opts[:user] || Etc.getlogin
```

maruware:
Thank for replying!

Your code is simple and better.
But a return value of Net::SSH::Config.for includes keys not in originally opts.
(ex. :auth_methods, :send_env)
If it's no problem, I will modify code at your suggestion.

ryotarai:
> But a return value of Net::SSH::Config.for includes keys not in originally opts.
> (ex. :auth_methods, :send_env)
> If it's no problem, I will modify code at your suggestion.

Yes, no problem!

maruware:
I have modified. Please check it.

ryotarai:
Thank you!

PR #117, opened 2015-03-31
Use given attribute value even if it's falsey

### LINKED ISSUE #108: from_image method
Using Docker backend, recipes and base image are inseparable in most cases.

### LINKED ISSUE #116: defined resources ignores falsey arguments
``` ruby
define :awesome, flag: 'default' do
  p ["(^-^)/", params[:flag]]
end

awesome 'nil' do
  flag nil
end

awesome 'false' do
  flag false
end

awesome 'true' do
  flag true
end
```

```
 INFO : Starting Itamae...
["(^-^)/", "default"]
["(^-^)/", "default"]
["(^-^)/", true]
 INFO : Recipe: /tmp/define.rb
```

sorah:
Fixes #116

### DIFF
--- a/lib/itamae/resource/base.rb
+++ b/lib/itamae/resource/base.rb
@@ -230,7 +230,7 @@ module Itamae
       def process_attributes
         self.class.defined_attributes.each_pair do |key, details|
           @attributes[key] ||= @resource_name if details[:default_name]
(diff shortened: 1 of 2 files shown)


sorah:
wrong link -- 108 → #116

ryotarai:
Thank you!

PR #123, opened 2015-04-25
add ohai_directory option.

kotakanbe:
pass the ohai plugin directory to use in recipe.


ryotarai:
Thank you for sending a PR, but I'm going to remove `--ohai` feature because currently Specinfra has [host inventory feature](http://serverspec.org/host_inventory.html).

Isn't the host inventory of Specinfra enough for you?

kotakanbe:
OK, I understatnd.
I will send PR to Specinfra if necessary.
Thank you.

PR #125, opened 2015-04-29
Fix a glitch on raising exception when source doesn't exist.

mozamimy:
I found a glitch when source file or directory doesn't exist.

For example, there is a itamae recipe file,

```ruby
remote_directory "/path/to/destination" do
  source "/path/to/source"
end
```

When `/path/to/source` does not exist, itamae says following exception message,
```
backend.rb:130:in `send_directory': uninitialized constant Itamae::Backend::Base::Error (NameError)
```

So I fixed the problem by using `SourceNotExistError`.


ryotarai:
Thank you!

ryotarai:
released v1.2.15

PR #126, opened 2015-04-29
Fix for disable_sudo option

evalphobia:
When using ssh with `--sudo=false` option, it results in error for `net-ssh`.
`disable_sudo` is not valid option in `net-ssh`, so move it from `ssh_options`.

### DIFF
--- a/spec/unit/lib/itamae/backend_spec.rb
+++ b/spec/unit/lib/itamae/backend_spec.rb
@@ -52,5 +52,43 @@ module Itamae
         end
       end
     end
+
+    describe Ssh do
+
+      describe "#ssh_options" do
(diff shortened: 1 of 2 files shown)


spec/unit/lib/itamae/backend_spec.rb
```diff
@@ -52,5 +52,43 @@ def initialize(_)
         end
       end
     end
+
+    describe Ssh do
+
```
evalphobia: actually this is first time for me to write rspec code, so it does not meet your requirement.
if so, rewrite it for the quality!

ryotarai:

ryotarai:
released v1.2.15

evalphobia:
appreciate your quick work! ;)

Issue #132, opened 2015-05-16
Supporting provider for service resource like chef


sonots:
I want to have `provider` attribute for `service` resource which chef has. 

In the chef, writing `service` resource like below allows us to change the service command to `upstart`. 
```ruby
service "example_service" do
  action :start
  provider Chef::Provider::Service::Upstart
  notifies :restart, "service[nginx]", :immediately
end
```

Also, with writing a gem like https://github.com/hirose31/chef-provider-service-daemontools, we add a provider `Chef::Provider::Service::Daemontools` to change the service commands to one of daemontools. 

I am willing to implement this, but I want to disscuss the implementation design beforehand. 

### 1) under way

Specinfra has some of `check_is_running_under_#{provider}` methods already like [this](https://github.com/serverspec/specinfra/blob/8978d69e5804978ff2252f20c5e8ec91ba25fe0c/lib/specinfra/command/base/service.rb#L11-L17), so one way would be to change `lib/itamae/resource/service.rb` as this patch > https://gist.github.com/sonots/0b601f610b04d2c27cdd. And, add `xxx_under_xxx` methods to specinfra. 

### 2) module way

Specinfra has `Specinfra::Command::Module::Systemd` at [lib/specinfra/command/module/systemd.rb](https://github.com/serverspec/specinfra/blob/d4a29586dc59540333e135fff07a0588994175db/lib/specinfra/command/module/systemd.rb), so one way would be to support to specify a module with `provider` attribute, and mixin it on calling `run_specinfra` by some ways. 

### 3) resource way

Giving up to provide `provider` attribute, and implement the feature as another `resource` like

```
daemontools_service "foo" do
  action :start
end
```

I think this design is heavily related with the implementation design in `specinfra`, so it is better to discuss with both @mizzy and @ryotarai. 

EDIT: I also want to have `provider` attribute for `package` resource like chef > https://docs.chef.io/resource_package.html. Chef allows us to specify `provider Chef::Provider::Package::Rpm`, also chef provides `rpm_package` **resource** in this case. 

Please let me know if there is a discussion already.

ryotarai:
@sonots Thank you for your proposal.
"1) under way" sounds good for me for now. Please send a pull request.
In the future, I may introduce general provider feature (not package resource specific) into Itamae.

sorah:
I want this feature too.

sonots:
[...]

PR #134, opened 2015-05-20
Support provider for service resource

sonots:
Fix https://github.com/itamae-kitchen/itamae/issues/132

Note that only `:init` and `:systemd` providers are fully implemented currently by specinfra. 

`daemontools` provider requires another parameter for `enable` action ([specinfra/daemontools.rb](https://github.com/serverspec/specinfra/blob/80700e73c0a251d0353a17d83b5fa7bd3429bf31/lib/specinfra/command/module/service/daemontools.rb#L14)), so I don't support it with `service` resource, but `daemontools_service` resource https://github.com/sonots/itamae-plugin-resource-daemontools_service can be used. 

PS. I could not find a suitable place to write tests, please let me know if there are.

### DIFF
--- a/lib/itamae/resource/service.rb
+++ b/lib/itamae/resource/service.rb
@@ -5,6 +5,12 @@ module Itamae
     class Service < Base
       define_attribute :action, default: :nothing
       define_attribute :name, type: String, default_name: true
+      define_attribute :provider, type: Symbol, default: nil
+
+      def initialize(*args)
+        super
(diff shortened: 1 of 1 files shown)


ryotarai:
Thank you for your PR.

> PS. I could not find a suitable place to write tests, please let me know if there are.

It's okay. I'll add tests...

ryotarai:
released in v1.2.17

sonots:
Thanks!

PR #135, opened 2015-06-02
Wrap host inventory value with Hashie::Mash.

ryotarai:
Make host inventory accessible by method calls like the other values.


ryotarai:
cc: @mizzy

mizzy:
I see.

PR #137, opened 2015-06-10
Show error message when specified action is unavailable in dry_run mode

eagletmt:
Because I often make a mistake in action name :trollface:


ryotarai:
Good catch. Thank you.

eagletmt:
:v:

PR #140, opened 2015-06-30
Add "recursive" attribute to git resource

mmasaki:
I propose an new attribute "recursive" into git resource.

Example:
```
git "foo" do
  repository "https://github.com/foo/including_submodules.git"
  recursive true
end
```

Then, "git clone --recursive https://github.com/foo/including_submodules.git" will be executed.

### DIFF
--- a/lib/itamae/resource/git.rb
+++ b/lib/itamae/resource/git.rb
@@ -30,11 +30,6 @@ module Itamae
         if run_specinfra(:check_file_is_directory, attributes.destination)
           run_command_in_repo(['git', 'fetch', 'origin'])
         else
-          if attributes.recursive
-            run_command(['git', 'clone', '--recursive', attributes.repository, attributes.destination])
-          else
-            run_command(['git', 'clone', attributes.repository, attributes.destination])
-          end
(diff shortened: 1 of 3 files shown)


ryotarai:
Great but I think the code is a little duplicated. How about the following?

```ruby
cmd = ['git', 'clone']
cmd << '--recursive' if attributes.recursive
cmd << attributes.repository << attributes.destination
run_command(cmd)
```

mmasaki:
Tests were failed, but I think it's not this change's sake.
Please consider re-testing and merging this change.

lib/itamae/resource/git.rb
```diff
@@ -29,7 +30,15 @@ def action_sync(options)
         if run_specinfra(:check_file_is_directory, attributes.destination)
           run_command_in_repo(['git', 'fetch', 'origin'])
         else
-          run_command(['git', 'clone', attributes.repository, attributes.destination])
+          if attributes.recursive
```
ryotarai: L33-37: Please remove these lines

mmasaki:
That's my mistake...

ryotarai:

PR #141, opened 2015-07-02
Add disable_sudo to each resource

sue445:
# Feature
disable sudo only in a specified resource.

It likes to `let(:disable_sudo) { true }`  in Serverspec

# Example
```sh
itamae ssh --sudo
```

```ruby
execute "whoami" do
  disable_sudo false
end
# => sudo whoami

execute "whoami" do
  disable_sudo true
end
# => whoami
```

### DIFF
--- a/lib/itamae/resource/base.rb
+++ b/lib/itamae/resource/base.rb
@@ -12,6 +12,7 @@ module Itamae
         attr_reader :verify_commands
         attr_reader :only_if_command
         attr_reader :not_if_command
+        attr_reader :temporary_disable_sudo
 
         def initialize(resource)
           @resource = resource
@@ -68,6 +69,10 @@ module Itamae
         def verify(command)
           @verify_commands << command
         end
+
+        def disable_sudo(temporary_disable_sudo)
+          @temporary_disable_sudo = temporary_disable_sudo
+        end
       end
 
       @defined_attributes ||= {}
@@ -114,6 +119,7 @@ module Itamae
           @only_if_command = context.only_if_command
           @not_if_command = context.not_if_command
           @verify_commands = context.verify_commands
+          @temporary_disable_sudo = context.temporary_disable_sudo
         end
 
         process_attributes
@@ -287,7 +293,18 @@ module Itamae
 
         args.last[:user] ||= attributes.user
 
-        backend.run_command(*args)
+        if @temporary_disable_sudo.nil?
+          backend.run_command(*args)
+        else
+          original_disable_sudo = Specinfra.configuration.disable_sudo
+          Specinfra.configuration.disable_sudo = @temporary_disable_sudo
+
+          result = backend.run_command(*args)
+
+          Specinfra.configuration.disable_sudo = original_disable_sudo
+
+          result
+        end
       end
 
       def check_command(*args)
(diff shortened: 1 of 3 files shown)


lib/itamae/resource/base.rb
```diff
@@ -287,7 +293,18 @@ def run_command(*args)
 
         args.last[:user] ||= attributes.user
 
-        backend.run_command(*args)
+        if @temporary_disable_sudo.nil?
+          backend.run_command(*args)
+        else
+          original_disable_sudo = Specinfra.configuration.disable_sudo
```
sue445: I want to do refactor, but I don't have good idea...
Do you have any idea?

ryotarai:
Itamae supports `--no-sudo` option to disable sudo. What situation do you need disabling sudo for each resource?

sue445:
@ryotarai 

> What situation do you need disabling sudo for each resource?

By default I want to use the `sudo`, but I don't want to use `sudo` in a particular resource

```ruby
    git "/usr/local/src/node" do
      repository  "https://github.com/joyent/node.git"
    end
```

then

```sh
sudo git clone https://github.com/joyent/node.git /usr/local/src/node
Cloning into '/usr/local/src/node'...
fatal: Unable to find remote helper for 'https'
```

So I don't want to use `sudo` only this `git` .

But, Sorry this is my problem :sweat_smile:
This problem is solved just now! (install `libcurl4-gnutls-dev` and rebuild `git`)

ryotarai:
@sue445 I understood your situation.
With `user` attribute, all commands in the resource will be executed by the `user`. (all Itamae resources have `user` attribute)
Does it work for you?

sue445:
> With user attribute, all commands in the resource will be executed by the user. 

Hmm..

`sudo -u` is not allowed :confounded:

Probably, this is problem of my company infrastructure 

```sh
$  sudo -u sueyoshi_go git clone https://github.com/joyent/node.git /usr/local/src/node

Sorry, user sueyoshi_go is not allowed to execute '/usr/local/bin/git clone https://github.com/joyent/node.git /usr/local/src/node' as sueyoshi_go on xxxxxxxx.
```

supermomonga:
+1 for this.
I can use `sudo -u` in my env but I think we still need this for some situations.

For example, I sometime want to clone private git repository via ssh protocol. I don't want put secret-key to the server, so I use `ForwardAgent`. It works fine without sudo but if I use `sudo -u`, it can't read ssh-agent's sock file so failed to pull private repository. (ref [sudo -u someuserした場合にsudoコマンドを打ったユーザーの秘密鍵の鍵を使いたい](http://ja.stackoverflow.com/questions/6256/sudo-u-someuser%E3%81%97%E3%81%9F%E5%A0%B4%E5%90%88%E3%81%ABsudo%E3%82%B3%E3%83%9E%E3%83%B3%E3%83%89%E3%82%92%E6%89%93%E3%81%A3%E3%81%9F%E3%83%A6%E3%83%BC%E3%82%B6%E3%83%BC%E3%81%AE%E7%A7%98%E5%AF%86%E9%8D%B5%E3%81%AE%E9%8D%B5%E3%82%92%E4%BD%BF%E3%81%84%E3%81%9F%E3%81%84))

So, if I think `diable_sudo` is very handy in some situations.

supermomonga:
@ryotarai Could you consider reopen this issue?

PR #142, opened 2015-07-08
docker command options are not enough

marcy-terui:
Currently, Itamae is ready to provision the running container.
But, it can't provision them.
Because the command options are not enough.

https://github.com/itamae-kitchen/itamae/blob/master/lib/itamae/backend.rb#L227-L230
https://github.com/serverspec/specinfra/blob/master/lib/specinfra/backend/docker.rb#L15-L22


ryotarai:
Thank you

PR #149, opened 2015-07-17
Allow array gem_binary

eagletmt:
Sometimes I'd like to set custom gem command like
`env RBENV_VERSION=2.2.2 rbenv exec gem` .


ryotarai:
Thank you!

PR #150, opened 2015-07-17
Check directory is empty or not for git resource

tacahilo:
Problem
---

- Although `/path/to/destination/` is an empty directory, it doesn't run `git clone`

Suggestion
---

- Check directory is empty or not.


ryotarai:
Thank you!

PR #156, opened 2015-08-18
Use `default.rb` if include_recipe is called with directory path.

### LINKED ISSUE #99: Use `default.rb` if include_recipe is called with directory path.
```ruby
include_recipe "cookbooks/nginx"
```

If `cookbooks/nginx` is a directory and `cookbooks/nginx/default.rb` exists, `default.rb` will be included.

Rudolph-Miller:
close #99

### DIFF
--- a/lib/itamae/recipe.rb
+++ b/lib/itamae/recipe.rb
@@ -106,8 +106,9 @@ module Itamae
       end
 
       def include_recipe(target)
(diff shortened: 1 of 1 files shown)


ryotarai:
Thank you!

Rudolph-Miller:
Thanks. :)

PR #165, opened 2015-09-23
Ignore yaml when the result is false

k0kubun:
Because YAML does not have a good way to write empty hash (I only know `YAML.load('{}') #=> {}`), I sometimes comment out all lines in node.yml.
But if node.yml has only comments, `YAML.load` returns false. It crashes with `hash.merge!(false)`.

```rb
YAML.load('# comment') #=> false
```

So I want to fallback `YAML.load` result to empty hash when it has only comments.


ryotarai:
Good catch. Thank you.

ryotarai:

PR #166, opened 2015-09-23
Allow `include_recipe` to omit rb extension

k0kubun:
Like ruby's built-in `require`, it would be convenient to abridge `.rb` extension.
With this change, you can include `foo.rb` recipe by both `include_recipe 'foo.rb'` and `include_recipe 'foo'`. So this is backward-compatible.


ryotarai:
Thank you!

ryotarai:

PR #167, opened 2015-09-23
Allow `load_recipes` to load plugin recipes directly

k0kubun:
If you want to apply only a plugin recipe, you have to write a recipes which has only `include_recipe 'foo::bar'`. But if you can load plugin recipes directly, you can easily try plugin recipes.

```bash
$ gem install itamae-plugin-recipe-foo
$ itamae local foo::bar
```


ryotarai:
Thank you!

ryotarai:

PR #169, opened 2015-10-07
Fix not send notifies when file edited

kurochan:
not sent notifies following example.

```
file "/path/to/file" do
  action :edit
  block do |content|
    content.gsub!("world", "Itamae")
  end
  notifies :run, "execute[echo 1 >> /tmp/file_edit_notifies]"
end
```


ryotarai:
Good catch! Thank you.

PR #170, opened 2015-10-15
Change default permission for files to 644

tamano:
Solves #113 .

I didn't do anything for directories because it already had permission 755.

### DIFF
--- a/lib/itamae/resource/file.rb
+++ b/lib/itamae/resource/file.rb
@@ -150,6 +150,7 @@ module Itamae
                 else
                   f = Tempfile.open('itamae')
                   f.write(attributes.content)
+                  f.chmod(0644)
                   f.close
                   f.path
                 end
@@ -163,4 +164,3 @@ module Itamae
     end
   end
 end
-
--- a/spec/integration/default_spec.rb
+++ b/spec/integration/default_spec.rb
@@ -41,6 +41,10 @@ describe file('/tmp/directory_never_exist1') do
   it { should_not be_directory }
 end
 
+describe file('/tmp/default_permission_dir') do
+  it { should be_mode 755 }
+end
+
 %w!/tmp/template /tmp/template_auto!.each do |f|
   describe file(f) do
     it { should be_file }
@@ -68,6 +72,10 @@ end
 
 describe file('/tmp/never_exist2') do
   it { should_not be_file }
+end
+
+describe file('/tmp/default_permission_file') do
+  it { should be_mode 644 }
 end
 
 describe file('/tmp/notifies') do
--- a/spec/integration/recipes/default.rb
+++ b/spec/integration/recipes/default.rb
@@ -129,6 +129,12 @@ directory "/tmp/directory_never_exist1" do
   action :delete
 end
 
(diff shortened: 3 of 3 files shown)


ryotarai:
Thank you.

My concern are:

- Is Chef's behavior the same as this PR or not?
- `umask` value will be ignored

What do you think about these point?

tamano:
Thank you for reviewing.

> Is Chef's behavior the same as this PR or not?

Yes it does. I've tested. Below gist is the log. 

https://gist.github.com/tamano/3b5e844ae3bbb0e09e93

> umask value will be ignored

I think this depends on policy. IMO it's ok because of compatibility with Chef (Chef also ignores umask; See gist above).

ryotarai:
Sorry for very late reply.
This PR changes only permission of a file, doesn't it?
Permission of a directory should be changed too

tamano:
Thank you for your comment.

> This PR changes only permission of a file, doesn't it?

Yes. As I mentioned before, directories already had default permission 755.
Could you confirm the gist below?

https://gist.github.com/tamano/3b5e844ae3bbb0e09e93

Please let me know if my testing way is poor or wrong. 
Thanks.

ryotarai:
Sorry for late reply.

> Yes. As I mentioned before, directories already had default permission 755.

I think umask also affects permission of a directory like the below and it should be 755 regardless of umask.

```
$ cat tmp/umask.rb
file "sample-file"
directory "sample-dir"
$ umask
022
$ b bin/itamae local tmp/umask.rb
 INFO : Starting Itamae...
 INFO : Recipe: /Users/ryota-arai/src/github.com/itamae-kitchen/itamae/tmp/umask.rb
 INFO :   file[sample-file] exist will change from 'false' to 'true'
 INFO :   directory[sample-dir] exist will change from 'false' to 'true'
$ ls -la sample-*
-rw-r--r--  1 ryota-arai  staff  0  3 15 10:00 sample-file

sample-dir:
total 0
drwxr-xr-x   2 ryota-arai  staff   68  3 15 10:00 .
drwxr-xr-x  22 ryota-arai  staff  748  3 15 10:00 ..
$ umask 266
$ umask
0266
$ rm sample-*
rm: sample-dir: is a directory
$ rmdir sample-dir
$ b bin/itamae local tmp/umask.rb
 INFO : Starting Itamae...
 INFO : Recipe: /Users/ryota-arai/src/github.com/itamae-kitchen/itamae/tmp/umask.rb
 INFO :   file[sample-file] exist will change from 'false' to 'true'
 INFO :   directory[sample-dir] exist will change from 'false' to 'true'
$ ls -la sample-*
-r--------  1 ryota-arai  staff  0  3 15 10:01 sample-file

sample-dir:
total 0
dr-x--x--x   2 ryota-arai  staff   68  3 15 10:01 .
drwxr-xr-x  22 ryota-arai  staff  748  3 15 10:01 ..
```

ryotarai:
> I think umask also affects permission of a directory like the below and it should be 755 regardless of umask.

How about this? I'm going to close this issue for now but feed free to reopen.

PR #179, opened 2015-12-06
Fallback to autoload resource plugin

k0kubun:
Like recipe plugin, I want to use resource plugin without calling `require`. If the resource plugin follows `bundle gem`'s naming convention, require path of `foo` resource should be `itamae/plugin/resource/#{foo}`.

This autoload does not happen if it is explicitly required, so this change is backword-compatible.


ryotarai:
Good catch! Thank you.

ryotarai:
released in v1.8.0

PR #180, opened 2015-12-07
Create http_request resource

hico-horiuchi:
Create [http_request resource](https://docs.chef.io/resource_http_request.html) for itamae.
This inherits file resource.
A example is shown in below.

```ruby
http_request "/tmp/itamae.html" do
  headers "User-Agent" => "Ruby/#{RUBY_VERSION}"
  url "http://itamae.kitchen/"
end
```


hico-horiuchi:
I apologize for many fixes.
Should I rebase these commits?

spec/integration/default_spec.rb
```diff
@@ -70,6 +70,16 @@
   it { should_not be_file }
 end
 
+describe file('/tmp/http_request.html') do
+  it { should be_file }
+  its(:content) { should match(/Itamae/) }
```
ryotarai: This looks assuming `http://itamae.kitchen` always return a body including `Itamae` but it can change.

spec/integration/default_spec.rb
```diff
@@ -70,6 +70,16 @@
...
+end
+
+describe file('/tmp/http_request_headers.html') do
+  it { should be_file }
+  its(:content) { should match(/Itamae/) }
```
ryotarai: This can't be a test for request header.

ryotarai: How about https://httpbin.org/get ?

spec/integration/default_spec.rb
ryotarai: How about https://httpbin.org/get?from=itamae ?

hico-horiuchi:
That's great!
I'll use it.

ryotarai:
Thank you for the fix. Would you rebase the commits?

hico-horiuchi:
I rebased the commits!

lib/itamae/resource/http_request.rb
```diff
@@ -0,0 +1,29 @@
...
+module Itamae
+  module Resource
+    class HttpRequest < File
+      UrlNotFoundError = Class.new(StandardError)
+
+      define_attribute :headers, type: Hash, default: {}
+      define_attribute :url, type: String
+
+      def pre_action
+        unless attributes.url
```
ryotarai: `define_attribute` accepts `required` option.

```ruby
define_attribute :url, type: String, required: true
```

Instead of L13-L15, please use this.

lib/itamae/resource/http_request.rb
```diff
@@ -0,0 +1,29 @@
...
+        end
+
+        attributes.content = open(attributes.url, attributes.headers).read
+
+        super
+      end
+
+      private
+
+      def content_file
```
ryotarai: This method is the same as `File#content_file` so this can be deleted

hico-horiuchi:
Thank you for reviewing.
I fixed them.

PR #184, opened 2015-12-13
Improve http_request resource to use net/http

hico-horiuchi:
Use `net/http` on http_request resource to available DELETE. OPTIONS, POST and PUT as `action` .
And `message` that is sent by the HTTP request can be set when POST and PUT.
(Those features are same as [http_request](https://docs.chef.io/resource_http_request.html#actions) of Chef.)

For example:

```ruby
http_request "/tmp/http_request_post.html" do
  action :post
  message "love=sushi"
  url "https://httpbin.org/post?from=itamae"
end
```

However, the test of `action :options` is not yet added.
([httpbin](https://httpbin.org) does not support OPTIONS.)

### DIFF
--- a/lib/itamae/resource/http_request.rb
+++ b/lib/itamae/resource/http_request.rb
@@ -1,18 +1,52 @@
 require 'itamae'
-require 'open-uri'
+require 'uri'
+require 'net/http'
 
 module Itamae
   module Resource
     class HttpRequest < File
       UrlNotFoundError = Class.new(StandardError)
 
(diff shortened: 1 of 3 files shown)


ryotarai:
@hico-horiuchi I fixed dependency problem by https://github.com/itamae-kitchen/itamae/commit/1284b84c41d1e19e862b57b491eaf4228e7f01a8 . Would you rebase this PR?

hico-horiuchi:
@ryotarai Thank you!

PR #186, opened 2015-12-21
Add options attribute to gem_package resource

hico-horiuchi:
Add `options` attribute to `gem_package`.
(This attribute is same as [gem_package](https://docs.chef.io/resource_gem_package.html#properties) of Chef.)

For example:

```ruby
gem_package 'bundler' do
  options ['--no-ri', '--no-rdoc']
end
```

### DIFF
--- a/spec/integration/recipes/default.rb
+++ b/spec/integration/recipes/default.rb
@@ -65,6 +65,10 @@ gem_package 'tzinfo' do
   version '1.2.2'
 end
 
+gem_package 'bundler' do
+  options ['--no-ri', '--no-rdoc']
+end
+
 ######
 
 execute "echo -n > /tmp/notifies"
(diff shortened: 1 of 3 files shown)


spec/integration/recipes/default.rb
```diff
@@ -65,6 +65,10 @@
   version '1.2.2'
 end
 
+gem_package 'bundler' do
+  options ['--no-ri', '--no-rdoc']
```
ryotarai: `--no-ri` and `--no-rdoc` are deprecated options. Please use `--no-document`

hico-horiuchi:
@ryotarai Installed ruby at test is `1.9.1` , so `--no-document` options is not supported.
Should I back the options to `--no-ri --no-rdoc` ?
Or would you update ruby `2.0` and up?

```
DEBUG :       Executing `gem install --no-document bundler`...
DEBUG :         stdout | ERROR:  While executing gem ... (OptionParser::InvalidOption)
DEBUG :         stdout |     invalid option: --no-document
```

ryotarai:
> Installed ruby at test is 1.9.1 , so --no-document option is not supported.
> Should I back the options to --no-ri --no-rdoc ?

For now, using `--no-ri --no-rdoc` in tests is better.

hico-horiuchi:
I understand and remove the commit 8cc54c504d61a8a20857703f3eee83548da57f34 .

ryotarai:
Thank you!

Issue #185, opened 2015-12-15
Bundler version error


ryotarai:
reported by @hico-horiuchi 

```
Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    bundler (~> 1.3)

    vagrant was resolved to 1.8.0.dev, which depends on
      bundler (<= 1.10.6, >= 1.5.2)

  Current Bundler version:
    bundler (1.11.0)
This Gemfile requires a different version of Bundler.
Perhaps you need to update Bundler by running `gem install bundler`?
```

ryotarai:
fixed by https://github.com/itamae-kitchen/itamae/commit/1284b84c41d1e19e862b57b491eaf4228e7f01a8

PR #188, opened 2015-12-27
Fix bundler version error

hfm:
Vagrant relies on some internal APIs in Bundler core. 
The command of `bundle
exec itamae ssh --vagrant` will conflict with Vagrant APIs.

Using `Bundler.with_clean_env` is necessary when using itamae with bundler.

* ref:
 * #185
 * https://github.com/mitchellh/vagrant/issues/5581#issuecomment-91889166


ryotarai:
Thank you!

PR #193, opened 2016-01-19
Fix to use specinfra to avoid calling direct command.

yuichiro-naito:
For FreeBSD, 'chown --reference' and 'chmod --reference' are not supported.
Itamae should use the interfaces which specinfra supports.

### DIFF
--- a/lib/itamae/resource/file.rb
+++ b/lib/itamae/resource/file.rb
@@ -94,15 +94,15 @@ module Itamae
         if attributes.mode
           run_specinfra(:change_file_mode, @temppath, attributes.mode)
         else
-          mode=run_specinfra(:get_file_mode, attributes.path).stdout.chomp
+          mode = run_specinfra(:get_file_mode, attributes.path).stdout.chomp
           run_specinfra(:change_file_mode, @temppath, mode)
         end
 
         if attributes.owner || attributes.group
           run_specinfra(:change_file_owner, @temppath, attributes.owner, attributes.group)
         else
-          owner=run_specinfra(:get_file_owner_user, attributes.path).stdout.chomp
-          group=run_specinfra(:get_file_owner_group, attributes.path).stdout.chomp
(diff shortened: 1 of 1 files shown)


lib/itamae/resource/file.rb
```diff
@@ -94,13 +94,17 @@ def action_edit(options)
         if attributes.mode
           run_specinfra(:change_file_mode, @temppath, attributes.mode)
         else
-          run_command(['chmod', '--reference', attributes.path, @temppath])
+          mode=run_specinfra(:get_file_mode, attributes.path).stdout.chomp
```
ryotarai: Would you insert spaces like `mode = run_specinfra`?

lib/itamae/resource/file.rb
```diff
@@ -94,13 +94,17 @@ def action_edit(options)
...
+          run_specinfra(:change_file_mode, @temppath, mode)
         end
 
         if attributes.owner || attributes.group
           run_specinfra(:change_file_owner, @temppath, attributes.owner, attributes.group)
         else
-          run_command(['chown', '--reference', attributes.path, @temppath])
+          owner=run_specinfra(:get_file_owner_user, attributes.path).stdout.chomp
```
ryotarai: ditto https://github.com/itamae-kitchen/itamae/pull/193/files#r50629883

lib/itamae/resource/file.rb
```diff
@@ -94,13 +94,17 @@ def action_edit(options)
...
+          group=run_specinfra(:get_file_owner_group, attributes.path).stdout.chomp
```
ryotarai: ditto https://github.com/itamae-kitchen/itamae/pull/193/files#r50629883

yuichiro-naito:
@ryotarai 
I have fixed spaces around '='.
Thank you.

ryotarai:
released v1.9.4

Issue #174, opened 2015-11-03
`template` should restore mtime if nothing changed


smlx:
`template` updates the mtime of a file, even if nothing changes in the file. This can cause warnings such as this from `postfix`:

    postfix/smtpd[19718]: warning: database /etc/aliases.db is older than source file /etc/aliases

It would be good if itamae could cache the mtime of the file, and restore it if the file was not actually modified.

ryotarai:
https://github.com/itamae-kitchen/itamae/issues/194 is released in v1.9.3

PR #194, opened 2016-01-19
Don't replace files when they will not be updated

KitaitiMakoto:
Hi,

Updating files only when their contents will be updated is nice idea. I wrote patch for that.
Need I write test? If yes, we need to run test before and after running Itamae and it look over the top to me. That's why this pull request doesn't include test.

Background of this patch is:
I defined `notifies` to execute `restorecon` to `file` and `template` resources. But current Itamae doesn't kick `notifies` if contents of files are not updated in spite of the files are replaced with files whose contents are the same and SELinux security contexts are changed. If this patch is applied, Itamae doesn't touch files if they will not be updated, so security context will not be changed.

Could you consider the request?
Thanks.

### DIFF
--- a/lib/itamae/resource/file.rb
+++ b/lib/itamae/resource/file.rb
@@ -80,7 +80,9 @@ module Itamae
             updated!
           end
 
-          run_specinfra(:move_file, @temppath, attributes.path)
+          if updated?
+            run_specinfra(:move_file, @temppath, attributes.path)
+          end
         end
       end
 


smlx:
This probably addresses #174

ryotarai:
Thank you!

KitaitiMakoto:
Thank you for merging.

ryotarai:
released in v1.9.3

PR #197, opened 2016-02-03
Show diff on edit action of file resource in dry-run mode.

### LINKED ISSUE #196: edit action of file resource should update the content even if dry-run mode
https://github.com/itamae-kitchen/itamae/blob/446f9e29972b58c9514c0e25c1fedf27c008606c/lib/itamae/resource/file.rb#L23-L27

When it is dry-run mode and the file doesn't exist, skip to update the content

ryotarai:
fixes #196


ryotarai:
released v1.9.4

PR #198, opened 2016-02-07
Delete 'deploy-old' branch forcibly in action_sync

kaihar4:
'git branch -d <branch>' deletes the fully merged branch only.
But 'deploy-old' isn't merged in its upstream branch.
Therefore, it can not be deleted.

### DIFF
--- a/lib/itamae/resource/git.rb
+++ b/lib/itamae/resource/git.rb
@@ -55,7 +55,7 @@ module Itamae
           run_command_in_repo(["git", "checkout", target, "-b", DEPLOY_BRANCH])
 
           if deploy_old_created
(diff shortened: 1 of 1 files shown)


ryotarai:
`deploy-old` branch is usually fully-merged. Would you tell me details of the situation?

kaihar4:
Sorry for taking long to write you back.

Sorry, I had a mistake.
This problem occurs only when the repository has been force updated.
Should I ignore this case?

ryotarai:
> Should I ignore this case

I think this case should be fixed manually because deleting a branch with `-D` may cause an incident like discarding commits that are not pushed yet...

PR #206, opened 2016-04-04
Introduce --detailed-exitcode option.

ryotarai:
It is useful for automation of Itamae.

- 0 - succeeded and no diff
- 1 - failed
- 2 - succeeded and there is a diff

### DIFF
--- a/lib/itamae/cli.rb
+++ b/lib/itamae/cli.rb
@@ -24,6 +24,7 @@ module Itamae
       option :shell, type: :string, default: "/bin/sh"
       option :ohai, type: :boolean, default: false, desc: "This option is DEPRECATED and will be unavailable."
       option :profile, type: :string, desc: "[EXPERIMENTAL] Save profiling data", banner: "PATH"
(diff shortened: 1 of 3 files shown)


lib/itamae/cli.rb
```diff
@@ -24,6 +24,7 @@ def self.define_exec_options
       option :shell, type: :string, default: "/bin/sh"
       option :ohai, type: :boolean, default: false, desc: "This option is DEPRECATED and will be unavailable."
       option :profile, type: :string, desc: "[EXPERIMENTAL] Save profiling data", banner: "PATH"
+      option :detailed_exitcode, type: :boolean, default: false, desc: "Exit code 0 means succeeded and no diff, 1 means errored, and 2 means succeeded and there is a diff"
```
kwilczynski: I would suggest to change the wording to (or similar):

0 - The run succeeded with no changes or failures.
1 - The run failed.
2 - The run succeeded, and some resources were changed.

As plain word "diff" is a bit of a misnomer.

kwilczynski:
This is excellent! Thank you!

lib/itamae/cli.rb
ryotarai: Thank you!

PR #209, opened 2016-04-11
Not surrounded by LF in the ANSI escape sequence.

daic-h:
When reading the output of the log from another program,
chomp of the new line does not work correctly.


ryotarai:
Thank you!

PR #211, opened 2016-04-30
Add `--ssh-config` option to `itamae ssh` to specify ssh_config

koshigoe:
I need `-F` option like SSH to specify alternative ssh_config file.

```
$ itamae ssh --ssh-config .ssh/development -h devhost -y nodes/web.yml roles/web.rb
```

### DIFF
--- a/lib/itamae/backend.rb
+++ b/lib/itamae/backend.rb
@@ -230,7 +230,8 @@ module Itamae
         opts[:host_name] = @options[:host]
 
         # from ssh-config
-        opts.merge!(Net::SSH::Config.for(@options[:host]))
+        ssh_config_files = @options[:ssh_config] ? [@options[:ssh_config]] : Net::SSH::Config.default_files
+        opts.merge!(Net::SSH::Config.for(@options[:host], ssh_config_files))
         opts[:user] = @options[:user] || opts[:user] || Etc.getlogin
         opts[:keys] = [@options[:key]] if @options[:key]
         opts[:port] = @options[:port] if @options[:port]
--- a/lib/itamae/cli.rb
+++ b/lib/itamae/cli.rb
@@ -43,6 +43,7 @@ module Itamae
     option :user, type: :string, aliases: ['-u']
     option :key, type: :string, aliases: ['-i']
     option :port, type: :numeric, aliases: ['-p']
+    option :ssh_config, type: :string, aliases: ['-F']
     option :vagrant, type: :boolean, default: false
     option :ask_password, type: :boolean, default: false
     option :sudo, type: :boolean, default: true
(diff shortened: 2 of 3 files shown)


ryotarai:
`--ssh-config` option looks good to me, but I think aliasing `-F` option is not good idea because it can be used for another meaning in the future

koshigoe:
>  but I think aliasing -F option is not good idea because it can be used for another meaning in the future

Yes, I think so too. But I have no idea.
Can I remove alias? (only `--ssh-option`)

koshigoe:
MEMO:

At wercker, pass `rake spec:unit`, but `Time-out happened`.

> Manual page bundle-exec line 1 (press h for help or q to quit)

why opened man page?

```
Vagrant appears to be running in a Bundler environment. Your 
existing Gemfile will be used. Vagrant will not auto-load any plugins
installed with `vagrant plugin`. Vagrant will autoload any plugins in
the 'plugins' group in your Gemfile. You can force Vagrant to take over
with VAGRANT_FORCE_BUNDLER.

You appear to be running Vagrant outside of the official installers.
Note that the installers are what ensure that Vagrant has all required
dependencies, and Vagrant assumes that these dependencies exist. By
running outside of the installer environment, Vagrant may not function
properly. To remove this warning, install Vagrant using one of the
official packages from vagrantup.com.

Bringing machine 'trusty' up with 'digital_ocean' provider...
==> trusty: Droplet is already active
["bundle", "exec", "bin/itamae", "ssh", "-h", "104.236.14.139", "-u", "root", "-p", "22", "-i", "/home/ubuntu/.ssh/id_rsa.vagrant", "-l", "debug", "-j", "spec/integration/recipes/node.json", "spec/integration/recipes/default.rb", "spec/integration/recipes/default2.rb", "spec/integration/recipes/redefine.rb"]
BUNDLE-EXEC(1)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          BUNDLE-EXEC(1)

NAME
       bundle-exec - Execute a command in the context of the bundle

SYNOPSIS
       bundle exec [--keep-file-descriptors] command

DESCRIPTION
       This command executes the command, making all gems specified in the Gemfile(5) available to require in Ruby programs.

       Essentially, if you would normally have run something like rspec spec/my_spec.rb, and you want to use the gems specified in the Gemfile(5) and installed via bundle install(1) bundle-install.1.html, you should run bundle exec rspec spec/my_spec.rb.

       Note that bundle exec does not require that an executable is available on your shell´s $PATH.

OPTIONS
       --keep-file-descriptors
              Exec in Ruby 2.0 began discarding non-standard file descriptors. When this flag is passed, exec will revert to the 1.9 behaviour of passing all file descriptors to the new process.

BUNDLE INSTALL --BINSTUBS
       If you use the --binstubs flag in bundle install(1) bundle-install.1.html, Bundler will automatically create a directory (which defaults to app_root/bin) containing all of the executables available from gems in the bundle.

       After using --binstubs, bin/rspec spec/my_spec.rb is identical to bundle exec rspec spec/my_spec.rb.
 Manual page bundle-exec line 1 (press h for help or q to quit)
```

ryotarai:
Sorry for late response. I'm going to merge this. Thank you

PR #213, opened 2016-06-02
Fix url in gemspec and doc

sue445:
`spec.homepage` is used in https://rubygems.org/gems/itamae

This is link to https://github.com/ryotarai/itamae 

By the way, codeclimate link in README.md is old
https://codeclimate.com/github/ryotarai/itamae is not been updated

i recommend to replace to new one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ $ bundle exec rake spec
 
 ## Contributing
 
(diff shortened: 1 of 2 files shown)


ryotarai:
Thank you!

sue445:
😄

PR #215, opened 2016-06-27
send_file failed against docker backend

Joe-noh:
There is a problem that sending file is failed when the backend is docker. Reproduciton code is [here](https://github.com/Joe-noh/cannot_send_file).

This PR fixes the problem. Docker backend needs second argument for `send_file` to be a directory path. but I'm not sure this changes are good solution for the problem. please give me an advice if you find any other way to solve it.

Thanks.


ryotarai:
Thank you!

PR #216, opened 2016-06-28
Add uninstall action to gem_package

sue445:
Support feature of `gem uninstall`

# Example
```ruby
# uninstall all versions of gem
gem_package "bundler" do
  action :uninstall
end

# uninstall a specific version of gem
gem_package "bundler" do
  action :uninstall
  version "1.12.5"
end
```

### DIFF
--- a/lib/itamae/resource/gem_package.rb
+++ b/lib/itamae/resource/gem_package.rb
@@ -46,10 +46,7 @@ module Itamae
       end
 
       def action_uninstall(action_options)
-        if current.installed
-          uninstall!
-          updated!
-        end
+        uninstall! if current.installed
       end
 
       def action_upgrade(action_options)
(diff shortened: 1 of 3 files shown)


lib/itamae/resource/gem_package.rb
```diff
@@ -43,6 +45,13 @@ def action_install(action_options)
         end
       end
 
+      def action_uninstall(action_options)
+        if current.installed
+          uninstall!
+          updated!
```
ryotarai: I think this line is not necessary because if `attributes.installed != current.installed`, Itamae marks the resource as updated automatically.

sue445: > Itamae marks the resource as updated automatically.

@ryotarai Did you mean this ?
https://github.com/itamae-kitchen/itamae/blob/169debf335001160644f534cebb5007a9587cc50/lib/itamae/resource/base.rb#L203-L206

ryotarai: Yes

sue445: I remove at b4f2254

ryotarai:
Thank you!

sue445:
😃

PR #217, opened 2016-06-29
Add support for login shell

mspaulding06:
I would like to have the option to use a login shell so that my path environment is set correctly.  There might be a better way to handle this.


ryotarai:
Thank you!

PR #218, opened 2016-07-10
Add recursive chmod/chown to remote_directory resource.

yamajun:
remote_directory resource with mode/owner/group attributes just only
work at root directory("path" attribute).

This patch add recursively option.  And set default(it is good or bad?).

Test case
---------

    remote_directory '/var/www/html/webapps' do
      owner 'apache'
      group 'apache'
      mode  'g+w'
      source webapps
    end

### DIFF
--- a/lib/itamae/resource/remote_directory.rb
+++ b/lib/itamae/resource/remote_directory.rb
@@ -9,6 +9,8 @@ module Itamae
       define_attribute :mode, type: String
       define_attribute :owner, type: String
       define_attribute :group, type: String
+      define_attribute :recursive_mode, type: [String, Symbol], default: :true
+      define_attribute :recursive_owner, type: [String, Symbol], default: :true
 
       def pre_action
         directory = ::File.expand_path(attributes.source, ::File.dirname(@recipe.path))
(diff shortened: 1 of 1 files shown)


lib/itamae/resource/remote_directory.rb
```diff
@@ -9,6 +9,8 @@ class RemoteDirectory < Base
       define_attribute :mode, type: String
       define_attribute :owner, type: String
       define_attribute :group, type: String
+      define_attribute :recursive_mode, type: [String, Symbol], default: :true
```
ryotarai: I think setting false as the default value is better.

ryotarai: Could you use `true` instead of symbol `:true`?

lib/itamae/resource/remote_directory.rb
```diff
@@ -56,10 +58,18 @@ def show_differences
 
       def action_create(options)
         if attributes.mode
-          run_specinfra(:change_file_mode, @temppath, attributes.mode)
+          if attributes.recursive_mode == :true
+            run_specinfra(:change_file_mode, @temppath, attributes.mode, :recursive => true)
```
ryotarai: If `attributes.recursive_mode` is boolean, you can simply write this `run_specinfra(:change_file_mode, @temppath, attributes.mode, recursive: attributes.recursive_mode)`

yamajun: Thanks to your suggestion!
Replace to your code and testing now.

lib/itamae/resource/remote_directory.rb
yamajun: I think better way of recursive chown/chmod in default in my local usage.
But, this patch changes behavior from current version.   It is not good.

I change to boolean false in default.

yamajun: I just change to default on false.
Sorry too late.

lib/itamae/resource/remote_directory.rb
```diff
@@ -9,6 +9,8 @@ class RemoteDirectory < Base
       define_attribute :mode, type: String
       define_attribute :owner, type: String
       define_attribute :group, type: String
+      define_attribute :recursive_mode, type: [TrueClass, FalseClass], default: false
+      define_attribute :recursive_owner, type: [TrueClass, FalseClass], default: false
```
ryotarai: From the name of `recursive_owner`, users may not expect that group will be also changed recursively.
How about change the name to `recursive_owner_and_group`?

PR #219, opened 2016-07-13
Add depth attribute to git resource

tyamagu2:
It would be nice to have depth option supported when you clone a git repository with long history.

This patch adds depth attribute to git resource that uses depth option when cloning like the example below;

```
git "/tmp/git_repo_depth_1" do
  repository "https://github.com/ryotarai/infrataster.git"
  depth 1
end

# => git clone --depth 1 https://github.com/ryotarai/infrataster.git /tmp/git_repo_depth_1
```


sorah:
Looks fine. @ryotarai 👓 ?

ryotarai:
LGTM
Thank you :+1:

ryotarai:
released in v1.9.10

Issue #223, opened 2016-09-30
Can't call task when use subscribes option


umisora:
## About
I have a problem.
Parent task don't called after children task executed, when I use subscribes option in Parent task.

## Environment
Host: CentOS release 6.3
Target: CentOS release 6.3
- ruby: ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-linux]
- itamae: 1.9.9

## TestCode
```
# test1/default.rb
include_recipe "../test2/default.rb"
execute 'task3' do
  action :nothing
  command 'echo "task3"'
  subscribes :run, 'execute[task4]'
end
```

```
# test2/default.rb
execute 'task4' do
    command 'echo "task4"'
end
```

## Result
```
# result
$ bundle exec itamae local cookbooks/test1/default.rb 
 INFO : Starting Itamae...
 INFO : Recipe: /Users/umisora/workspace/sys_setup_test/cookbooks/test1/default.rb
 INFO :   Recipe: /Users/umisora/workspace/sys_setup_test/cookbooks/test2/default.rb
 INFO :     execute[task4] executed will change from 'false' to 'true'
```
execute task is 1 task. Only Task4.
I want to execute Task3 after Task4 executed.

## Subscribes option success case

if test2 include test1, subscribes option is work.

```
$ bundle exec itamae local cookbooks/test2/default.rb 
 INFO : Starting Itamae...
 INFO : Recipe: /Users/umisora/workspace/sys_setup_test/cookbooks/test2/default.rb
 INFO :   Recipe: /Users/umisora/workspace/sys_setup_test/cookbooks/test1/default.rb
 INFO :   execute[task4] executed will change from 'false' to 'true'
 INFO :   Notifying run to execute resource 'task3' (delayed)
 INFO :   (because it subscribes this resource)
 INFO :   execute[task3] executed will change from 'false' to 'true'
```

ryotarai:
This issue is fixed by https://github.com/itamae-kitchen/itamae/pull/224
@umisora Would you check if this issue is fixed properly or not? (note that #224 is not released yet)

umisora:
@ryotarai 
Thank you so much for your support.
I was able to confirm that it has success.
LGTM!

```
$ bundle exec itamae docker cookbooks/test1/default.rb --image=centos:centos6
 INFO : Starting Itamae...
 INFO : Recipe: /Users/umisora/workspace/sys_setup_test/cookbooks/test1/default.rb
[...]

PR #224, opened 2016-10-02
Run a resource subscribing a resource in child resource.

ryotarai:
https://github.com/itamae-kitchen/itamae/issues/223


ryotarai:
released in v1.9.10

PR #227, opened 2016-10-18
Add support password authentication for ssh

xerespm:
Added support password authentication for ssh.

config.yml:

``` yaml
password: P@ssw0rd
```

and do as below.

``` shell
$ itamae ssh -h <host> -u <user> -c config.yml recipe.rb
```

You can run `itamae ssh` by password authentication.


ryotarai:
released in v1.9.10

PR #228, opened 2016-10-26
Change file owner first, then change file permissions.

EricDuminil:
suid and sticky bits are ignored on Linux otherwise.


sorah:
This patch makes sense. Thank you!

EricDuminil:
My pleasure!

ryotarai:
released in v1.9.10

PR #231, opened 2016-11-17
Add Option. docker_container_create_options

umisora:
# About
You can set docker_container_create_options to config.yaml.

# Usecase
`docker_container_create_options` 
I want to set hostname to container when exec itamae.
My Itamae-Recipes has Case Statement by host name.

Is this option will merge, I will be able to set hostname to container.. 

# options
Do not prepare option that can be passed as arguments.
You can set option to `config.yaml` when using `--config` arguments.

example)

itamae.config.yaml
```
docker_container_create_options:
  name: itamae-docker
  Hostname: itamae-docker
  Domainname: localhost.com
```

### DIFF
--- a/lib/itamae/backend.rb
+++ b/lib/itamae/backend.rb
@@ -288,6 +288,7 @@ module Itamae
           docker_image: @options[:image],
           docker_container: @options[:container],
           shell: @options[:shell],
+          docker_container_create_options: @options[:docker_container_create_options],
         )
       end
     end


ryotarai:
```
  Hostname: itamae-docker
  Domainname: localhost.com
```

These capitalized keys are not comfortable for me in terms of consistency.

umisora:
[Docker API](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#/tag-an-image-into-a-repository%23create-a-container) reference uses camelcase strings. So I wrote example `itamae.config.yaml` with camelcase strings.

Do you think it is better to insert some components to translate `itamae.config.yaml` to fit with API's standard formats (in this case to translate to camelcase strings) ?

Do you agree to make it possible to set the hostname with the option of [docker_container_create_options]?

umisora:
and more...
Even If you use Lower case in `itamae.config.yaml`,  it works.

umisora:
|ω・`） @ryotarai

ryotarai:
> Even If you use Lower case in itamae.config.yaml, it works

Okay, I gonna merge this.

umisora:
Thank you !

ryotarai:
released in v1.9.11

Issue #232, opened 2016-12-01
ssh with multiple keys fails to use the right one


torras:
Even with all the keys configured in the user's .ssh/config file, added to the ssh agent and the right private key file selected by command line, net-ssh fails to select the right key.

example run:

from the development machine, with key `dd:20:8d:5d:27:df:a2:bb:ae:6d:11:55:40:b6:a7:b0`:
(id from `~$ ssh-keygen -l -E md5 -f ~/.ssh/id_rsa.pub`)

```bash
~/itamae_project$ itamae ssh -y nodes/test.yml -h test -u user -i ~/.ssh/id_rsa recipe.rb
 INFO : Starting Itamae...
 INFO : Loading node data from /home/rhialto/src/wape/itamae_infra/nodes/test.yml...
/home/user/.rvm/gems/ruby-2.3.1@itamae/gems/net-ssh-3.2.0/lib/net/ssh/transport/session.rb:198:in `block in poll_message': disconnected: Too many authentication failures for user from 49.49.49.49 port 33208 ssh2 (2) (Net::SSH::Disconnect)
```

server's ssh daemon log in debug mode from the same run:
```bash
... not interesting ...
sshd[3578]: Connection from 49.49.49.49 port 33208 on 192.168.17.4 port 22
...
sshd[3578]: Failed publickey for user from 49.49.49.49 port 33208 ssh2: RSA fd:ec:7a:29:ea:d3:48:7b:d7:4a:77:d2:bf:8f:09:fc
...
sshd[3578]: Failed publickey for user from 49.49.49.49 port 33208 ssh2: RSA eb:e3:42:1e:25:3a:cc:f1:50:ab:99:69:9d:dd:c2:bc
...
sshd[3578]: Failed publickey for user from 49.49.49.49 port 33208 ssh2: RSA a7:99:7e:1e:ce:74:bd:e3:d7:f9:cf:be:e3:88:af:06
...
sshd[3578]: Failed publickey for user from 49.49.49.49 port 33208 ssh2: RSA 86:6e:d6:de:e5:94:af:90:28:ef:a7:bd:37:95:de:e5
...
sshd[3578]: Failed publickey for user from 49.49.49.49 port 33208 ssh2: RSA ed:38:c0:f3:42:47:60:41:4c:97:8a:9b:a1:b2:6d:f1
...
[...]

PR #233, opened 2016-12-01
added ssh option 'keys_only: true' when not using password login issue #232

torras:
From https://net-ssh.github.io/ssh/v2/api/classes/Net/SSH.html

:keys_only => set to true to use only private keys from keys and key_data parameters, even if ssh-agent offers more identities. This option is intended for situations where ssh-agent offers many different identites.

### DIFF
--- a/lib/itamae/backend.rb
+++ b/lib/itamae/backend.rb
@@ -257,6 +257,8 @@ module Itamae
           password = STDIN.noecho(&:gets).strip
           print "\n"
           opts.merge!(password: password)
+        else\
+          opts.merge!(keys_only: true)
         end
 
         opts


ryotarai:
ref: https://github.com/itamae-kitchen/itamae/issues/232

lib/itamae/backend.rb
```diff
@@ -257,6 +257,8 @@ def ssh_options
           password = STDIN.noecho(&:gets).strip
           print "\n"
           opts.merge!(password: password)
+        else
+          opts.merge!(keys_only: true)
```
ryotarai: Please set `keys_only: true` only if `@options[:key]` is set.

torras: Hi, @ryotarai, this is going to work when setting the key via command line, but is going to fail when the key is set by ssh config. Is this the intended behaviour?

lib/itamae/backend.rb
```diff
@@ -257,6 +257,8 @@ def ssh_options
           password = STDIN.noecho(&:gets).strip
           print "\n"
           opts.merge!(password: password)
+        else
+          opts.merge!(keys_only: true) if @options[:key]
```
ryotarai: [nits] `opts[:keys]` is better than `@options[:key]` in this context

PR #236, opened 2017-01-16
Fix problem of Windows CR-LF

urara4:
Fixed the problem that LF is replaced to CR-LF when file resource is used from windows.

### DIFF
--- a/lib/itamae/resource/file.rb
+++ b/lib/itamae/resource/file.rb
@@ -166,7 +166,7 @@ module Itamae
           src = if content_file
                   content_file
                 else
-                  f = Tempfile.open('itamae')
+                  f = Tempfile.open('itamae', :mode=>IO::RDWR | IO::CREAT | IO::TRUNC | IO::BINARY)
                   f.write(attributes.content)
                   f.close
                   f.path


lib/itamae/resource/file.rb
```diff
@@ -166,7 +166,7 @@ def send_tempfile
           src = if content_file
                   content_file
                 else
-                  f = Tempfile.open('itamae')
+                  f = Tempfile.open('itamae', :mode=>IO::RDWR | IO::CREAT | IO::TRUNC | IO::BINARY)
```
ryotarai: 1. I think this behavior should be optional, not by default.
2. Is not specifying only `IO::BINARY` sufficient for this purpose?

tomothy83:
I am a newbie of itamae, but a fan of this wonderful tool.

I was provisioning Linux nodes from Windows machine by `itamae ssh`, and got this problem in `template` resources of shell scripts.
Code change above also fixed my problem. Thanks!

The default open mode in `Tempfile.open` is "w+", which means when itamae executed in Windows machine, `LF` is converted to `CRLF`. This is not desirable in any case, I think.
In this script, the open mode should be `"w+b"`.  This "b" flag has no affect on non-DOS/Windows system. (`"w+b"` is the same as `IO::RDWR | IO::CREAT | IO::TRUNC | IO::BINARY`, right?)

In some case (such as provisioning configuration file into Windows machine), this conversion might be nice. Even though the need of conversion should be detected from the platform of PROVISIONED nodes, not from the platform of PROVISIONING machine like this.

Issue #243, opened 2017-07-28
Print Hashie warnings


hnmx4:
When node has key that conflict with a built-in method Hashie::Mash, print warning as the following.
Especially using `--ohai` option prints WARN about 50 lines.
```
W, [2017-07-28T07:00:30.573130 #3044]  WARN -- : You are setting a key that conflicts with a built-in method Hashie::Mash#size defined in Hash. This can cause unexpected behavior when accessing the key via as a property. You can still access the key via the #[] method.
```

unasuke:
Fixed

PR #244, opened 2017-07-28
Disable Mash warnings

hnmx4:
for https://github.com/itamae-kitchen/itamae/issues/243.
I use a Hashie developer’s solution that solves issue similar to issue#243 as  reference.

https://github.com/intridea/hashie/issues/394
> I just released Hashie v3.5.2 which has the disable_warnings feature for the Mash logger and sets the Hashie logger to the Rails logger when we detect that we're running in Rails. Hopefully, this helps!

https://github.com/intridea/hashie
> You can also disable the logging in subclasses of Mash:
>
> ```
> class Response < Hashie::Mash
>   disable_warnings
> end
> ```


hnmx4:
@unasuke hi. Congratulations on appointment to Itamae committee :tada: . I merged the master branch to this branch. Could you merge this branch?

lib/itamae/node.rb
```diff
@@ -1,4 +1,5 @@
 require 'itamae'
+require 'itamae/mash'
```
sue445: I prefer add `require 'itamae/mash'` to [lib/itamae.rb](https://github.com/itamae-kitchen/itamae/blob/v1.9.11/lib/itamae.rb) than here.

@unasuke Do you think?

unasuke:
@hnmx4 Thank you for catch-up! But still printing warn message on CI. :cry:  Could you fix it?

https://travis-ci.org/itamae-kitchen/itamae/jobs/439940863

lib/itamae/node.rb
unasuke: I agree!

hnmx4:
@unasuke Thank you for reply. Fix later. 🙃

sue445:
@hnmx4 ping

unasuke:
It merged in #273 . Thank you @hnmx4 ! :heart_eyes:

Issue #242, opened 2017-06-19
Test failed


pataiji:
Caused by power-assert gem using `%i` literal that isn't supported by ruby 1.9.

https://github.com/k-tsj/power_assert/blob/master/lib/power_assert/parser.rb#L60

```
 INFO :       gem_package[test-unit] installed will change from 'false' to 'true'
DEBUG :       Executing `gem install -v 3.2.0 test-unit`...
DEBUG :         stdout | Fetching: power_assert-1.0.2.gemFetching: power_assert-1.0.2.gem ( 79%)Fetching: power_assert-1.0.2.gem ( 88%)Fetching: power_assert-1.0.2.gem (100%)Fetching: power_assert-1.0.2.gem (100%)
DEBUG :         stdout | Fetching: test-unit-3.2.0.gemFetching: test-unit-3.2.0.gem (  4%)Fetching: test-unit-3.2.0.gem ( 13%)Fetching: test-unit-3.2.0.gem ( 17%)Fetching: test-unit-3.2.0.gem ( 20%)Fetching: test-unit-3.2.0.gem ( 25%)Fetching: test-unit-3.2.0.gem ( 29%)Fetching: test-unit-3.2.0.gem ( 34%)Fetching: test-unit-3.2.0.gem ( 41%)Fetching: test-unit-3.2.0.gem ( 44%)Fetching: test-unit-3.2.0.gem ( 51%)Fetching: test-unit-3.2.0.gem ( 57%)Fetching: test-unit-3.2.0.gem ( 65%)Fetching: test-unit-3.2.0.gem ( 72%)Fetching: test-unit-3.2.0.gem ( 73%)Fetching: test-unit-3.2.0.gem ( 81%)Fetching: test-unit-3.2.0.gem ( 89%)Fetching: test-unit-3.2.0.gem ( 95%)Fetching: test-unit-3.2.0.gem ( 98%)Fetching: test-unit-3.2.0.gem (100%)Fetching: test-unit-3.2.0.gem (100%)
DEBUG :         stdout | Successfully installed power_assert-1.0.2
DEBUG :         stdout | Successfully installed test-unit-3.2.0
DEBUG :         stdout | 2 gems installed
DEBUG :         stdout | Installing ri documentation for power_assert-1.0.2...
DEBUG :         stdout | 
DEBUG :         stdout | RDoc::Parser::Ruby failure around line 60 of
DEBUG :         stdout | lib/power_assert/parser.rb
[...]

PR #245, opened 2017-08-17
add support of FreeBSD Jail

yuichiro-naito:
I wrote a patch that makes Itamae to support FreeBSD jail (*1).
I tested all Itamae resources work with Specinfra-2.70.1.

My fix is quite simple.
 
 1. added jexec backend which executes specinfra commands in jail.
 2. added Itamae jail subcommand.

Here is a example to use Itamae jail subcommand.

```
$ su
# itamae jail --jail_name <jail_name> <recipe> ...
```

Itamae jail subcommand requires root privilege,
because FreeBSD's jexec command requires root privilege.

Will you see my fixes and merge this Pull Request?

*1: FreeBSD jail information
https://www.freebsd.org/doc/handbook/jails.html
https://ja.wikipedia.org/wiki/FreeBSD_jail

### DIFF
--- a/lib/itamae/backend.rb
+++ b/lib/itamae/backend.rb
@@ -212,6 +212,17 @@ module Itamae
       end
     end
 
+    class Jexec < Base
+      private
+      def create_specinfra_backend
+        Specinfra::Backend::Jexec.new(
+          shell: @options[:shell],
+          login_shell: @options[:login_shell],
+          jail_name: @options[:jail_name],
+        )
+      end
+    end
+
     class Ssh < Base
       private
       def create_specinfra_backend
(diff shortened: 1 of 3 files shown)


yuichiro-naito:
My approach doesn't change the test result.
I'm not sure why test fails.

hnmx4:
Probably, I think that the master branch test has failed. ref:https://github.com/itamae-kitchen/itamae/issues/242

yuichiro-naito:
https://github.com/itamae-kitchen/itamae/pull/247 seems to fix the test problem.
I will close this Pull Request and reopen new one.

PR #256, opened 2018-05-30
Fix spec

sue445:
Spec is failed because response format of https://httpbin.org/ was changed

```bash
$ curl -s "https://httpbin.org/get?from=itamae"
{"args":{"from":"itamae"},"headers":{"Accept":"*/*","Connection":"close","Host":"httpbin.org","User-Agent":"curl/7.54.0"},"origin":"183.77.207.223","url":"https://httpbin.org/get?from=itamae"}
```

### DIFF
--- a/spec/integration/default_spec.rb
+++ b/spec/integration/default_spec.rb
@@ -79,38 +79,38 @@ end
 
 describe file('/tmp/http_request.html') do
   it { should be_file }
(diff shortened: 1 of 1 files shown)


sue445:
Wercker build is failed, But my TravisCI build is successful (including #248, #256, #257 and #258 )

https://travis-ci.org/sue445/itamae/builds/385570279

c.f. #258

PR #273, opened 2018-12-14
Disable mash warnings (review catch up)

unasuke:
## original pull request description ( by @hnmx4 )
for https://github.com/itamae-kitchen/itamae/issues/243.
I use a Hashie developer’s solution that solves issue similar to issue#243 as  reference.

https://github.com/intridea/hashie/issues/394
> I just released Hashie v3.5.2 which has the disable_warnings feature for the Mash logger and sets the Hashie logger to the Rails logger when we detect that we're running in Rails. Hopefully, this helps!

https://github.com/intridea/hashie
> You can also disable the logging in subclasses of Mash:
>
> ```
> class Response < Hashie::Mash
>   disable_warnings
> end
> ```

------

I added commits for @sue445 's comment 

>  I prefer add require 'itamae/mash' to lib/itamae.rb than here.

https://github.com/itamae-kitchen/itamae/pull/244#discussion_r224300318

### DIFF
--- a/lib/itamae.rb
+++ b/lib/itamae.rb
@@ -13,6 +13,7 @@ require "itamae/notification"
 require "itamae/definition"
 require "itamae/ext"
 require "itamae/generators"
+require "itamae/mash"
 
 module Itamae
   # Your code goes here...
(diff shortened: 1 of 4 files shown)


unasuke:
YAY!

PR #278, opened 2018-12-27
Clean docker image for integration test if the test is green

pocke:
Problem
===

After run the integration test, the docker container for integration test is not dead. So we need to kill the container manually to re-run the integration test.

```sh
$ bundle exec rake spec:integration:all
docker run --privileged -d --name itamae ubuntu:trusty /sbin/init
docker: Error response from daemon: Conflict. The container name "/itamae" is already in use by container "2406eb53822694ce14d8b38f3f1736b7fdd96020cd59e23a11bdbdb5d47ea9b8". You have to remove (or rename) that container to be able to reuse that name.
See 'docker run --help'.
rake aborted!
Command failed with status (125): [docker run --privileged -d --name itamae u...]
/home/pocke/ghq/github.com/itamae-kitchen/itamae/Rakefile:28:in `block (5 levels) in <top (required)>'
/home/pocke/.rbenv/versions/trunk/bin/bundle:23:in `load'
/home/pocke/.rbenv/versions/trunk/bin/bundle:23:in `<main>'
Tasks: TOP => spec:integration:all => spec:integration:ubuntu:trusty => spec:integration:docker:ubuntu:trusty
(See full trace by running task with --trace)
```

Solution
===

Clean the docker container after integration test. And it cleans only if the test is green.
I think we need the docker container for debugging when the test case failed.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -14,17 +14,18 @@ namespace :spec do
 
   namespace :integration do
     targets = ["ubuntu:trusty"]
+    container_name = 'itamae'
 
     task :all     => targets
 
     targets.each do |target|
       desc "Run provision and specs to #{target}"
-      task target => ["docker:#{target}", "provision:#{target}", "serverspec:#{target}"]
+      task target => ["docker:#{target}", "provision:#{target}", "serverspec:#{target}", 'clean_docker_container']
 
       namespace :docker do
         desc "Run docker for #{target}"
         task target do
-          sh "docker run --privileged -d --name itamae #{target} /sbin/init"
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -14,17 +14,18 @@ namespace :spec do
 
   namespace :integration do
     targets = ["ubuntu:trusty"]
+    container_name = 'itamae'
```
unasuke: I think that name is more good if unique using `SecureRandom` .

Like `"itamae-#{SecureRandom.hex(4)}"`

What do you think that @pocke ? :blush:

sue445: I agree to @pocke

PR #281, opened 2019-01-13
Add integration test with `itamae local` command

pocke:
Problem
===

We don't have any integration test for `itamae local` command, so we cannot test #277.

Solution
====

Add an integation test for `itamae local` command.

Strategy:

1. Start a docker container for the test.
1. Execute `itamae local` comamnd in the container.
1. Execute the serverspec from the host machine.

Goal of this pull request
==

- [x] Add integration test script.
- [x] Run this test on CI
- [x] Make the test cases green with `itamae local` also
- [ ] Add test case for #277 (maybe I should work this todo on #277 )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: ruby
 sudo: required
+dist: xenial
 services:
   - docker
 rvm:
--- a/Rakefile
+++ b/Rakefile
@@ -20,7 +20,7 @@ namespace :spec do
     targets = ["ubuntu:trusty"]
     container_name = 'itamae'
 
-    task :all     => targets
+    task :all     => targets + ['spec:integration:local']
 
     targets.each do |target|
       desc "Run provision and specs to #{target}"
--- a/Rakefile
+++ b/Rakefile
@@ -69,7 +69,7 @@ namespace :spec do
         RSpec::Core::RakeTask.new(target.to_sym) do |t|
           ENV['DOCKER_CONTAINER'] = container_name
           t.ruby_opts = '-I ./spec/integration'
-          t.pattern = "spec/integration/*_spec.rb"
+          t.pattern = "spec/integration/[default|docker]_spec.rb"
         end
       end
 
--- a/Rakefile
+++ b/Rakefile
@@ -3,6 +3,10 @@ require 'rspec/core/rake_task'
 require 'tempfile'
 require 'net/ssh'
 
+Dir['tasks/*.rb'].each do |file|
+  require_relative file
+end
+
 desc 'Run unit and integration specs.'
 task :spec => ['spec:unit', 'spec:integration:all']
 
--- a/Rakefile
+++ b/Rakefile
@@ -41,6 +41,7 @@ namespace :spec do
               "spec/integration/recipes/default.rb",
               "spec/integration/recipes/default2.rb",
               "spec/integration/recipes/redefine.rb",
+              "spec/integration/recipes/docker.rb",
             ],
             [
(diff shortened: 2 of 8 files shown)


spec/integration/default_spec.rb
```diff
@@ -206,7 +178,7 @@
 end
 
 describe command('gem list') do
-  its(:stdout) { should include('rake (11.1.0)') }
+  its(:stdout) { should match(/^rake \(.*11.1.0.*\)/) }
```
pocke: Another rake version is installed in the docker image that is used for `itamae local` command test, so relax the test.

spec/integration/recipes/local.rb
```diff
@@ -0,0 +1,10 @@
+package 'sl' do
+  version '3.03-17+b2'
```
pocke: `ruby` docker image does not have `sl` package version `3.03-17`, it only has `3.03-17+b2`.

spec/integration/recipes/local.rb
```diff
@@ -0,0 +1,10 @@
...
+end
+
+######
+
+gem_package 'bundler' do
+  version '1.17.3'
+  options ['--no-document']
```
pocke: New rubygems, which is installed in `ruby` docker image, does not have `--no-rdoc` and `--no-ri` options. So use `--no-document` option instead.

Note: Ubuntu trusty, which is used in the existing integration test, installs Ruby 1.9.3!!!!! Rubygems on Ruby 1.9.3 has `--no-rodc` and `--no-ri` options :upside_down_face: .

.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: ruby
 sudo: required
+dist: xenial
```
pocke: Need upgrading ubuntu version to use new version docker.
In trusty, `docker exec` does not have `--workdir` option.

tasks/integration_local_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    docker_run
+    prepare
+    provision
+    serverspec
+    clean_docker_container
+  end
+
+  def docker_run
+    mount_dir = Pathname(__dir__).join('../').to_s
+    sh 'docker', 'run', '--privileged', '-d', '--name', CONTAINER_NAME, '-v', "#{mount_dir}:/itamae", "ruby:#{@ruby_version}", 'sleep', '1d'
```
pocke: `sleep` command is a hack...
`/sbin/init` is used for the existing integration test, but the image does not have `/sbin/init`.
So use `sleep` command instead.

pocke:
I think this pull request is ready to review. Can you review this pull request?

I still have an unchecked TODO, but I will work on the todo on #277 .

Rakefile
```diff
@@ -64,7 +69,7 @@ namespace :spec do
         RSpec::Core::RakeTask.new(target.to_sym) do |t|
           ENV['DOCKER_CONTAINER'] = container_name
           t.ruby_opts = '-I ./spec/integration'
-          t.pattern = "spec/integration/*_spec.rb"
+          t.pattern = "spec/integration/[default|docker]_spec.rb"
```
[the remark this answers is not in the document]
pocke: Currently, it is unnecessary, but I think it is necessary in the near future.
Because I'll add a test case(s) for #277 , and serverspec file. Probably the new serverspec file is needed only by the new test. Test for `itamae docker` should not load the new serverspec file, so it's necessary.

sue445:
@pocke Can I merge this PR?

PR #284, opened 2019-01-21
Suppress Ruby warnings

pocke:
With `RUBYOPT=-w`, Ruby reports some warnings of Itamae.
This pull request will suppress two warnings.

* circular require  aea48c1
* Uninitialized instance variable 824dd1e

Please read each commit message to know what the commit changes.

Note: Ruby still warns with this pull request, but the warnings are for dependencies (e.g. specinfra). So we should fix the dependencies to avoid all warnings.

### DIFF
--- a/lib/itamae/definition.rb
+++ b/lib/itamae/definition.rb
@@ -1,5 +1,3 @@
-require 'itamae'
-
 module Itamae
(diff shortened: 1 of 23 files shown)


k0kubun:
:+1:

PR #288, opened 2019-05-10
Refine `itamae docker`'s created message

pocke:
This change will make `itamae docker` command to display tag name after the image is created if `tag` option is specified.

For example:

```
$ bundle exec itamae docker --image base-image path/to/role.rb --tag awesome-something:latest

(snip)

 INFO : Image created: sha256:24b0ede58da84ed28bafb629121d16b87e2d36cb96d8361413fdb2555876cac8, and tagged as awesome-something:latest
```

I think displaying created tag name is useful. It remind me the tag name.
And it is the same behaviour with `docker build` command.

```
$ docker build -t foo:bar .

(snip)

Successfully built 24b0ede58da8
Successfully tagged foo:bar
```

### DIFF
--- a/lib/itamae/backend.rb
+++ b/lib/itamae/backend.rb
@@ -292,7 +292,9 @@ module Itamae
         /\A(?<repo>.+?)(?:|:(?<tag>[^:]+))\z/.match(@options[:tag]) do |m|
           image.tag(repo: m[:repo], tag: m[:tag])
         end
-        Itamae.logger.info "Image created: #{image.id}"
+        log_message = "Image created: #{image.id}"
+        log_message << ", and tagged as #{@options[:tag]}"
+        Itamae.logger.info log_message
       end
 
       private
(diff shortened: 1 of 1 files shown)


lib/itamae/backend.rb
```diff
@@ -292,7 +292,9 @@ def finalize
         /\A(?<repo>.+?)(?:|:(?<tag>[^:]+))\z/.match(@options[:tag]) do |m|
           image.tag(repo: m[:repo], tag: m[:tag])
         end
-        Itamae.logger.info "Image created: #{image.id}"
+        log_message = "Image created: #{image.id}"
+        log_message << ", and tagged as #{@options[:tag]}"
```
unasuke: I got this log message in not passed `--tag` option.

```
INFO : Image created: sha256:5647ea2e1d250a3fd526ee434c4ae3bcbaabd9797bbbcba9077bab39a759e0a7, and tagged as
```

I think more better to add guard clause to log message build phase. :wink: 

```suggestion
        log_message << ", and tagged as #{@options[:tag]}" if @options[:tag]
```

pocke: Thank you! :+1: :+1: It makes sence absolutely.
I was going to write the conditon, but it looks I forgot the condition. :innocent:

PR #290, opened 2019-05-16
Add lib/itamae to $LOAD_PATH in bin/itamae script

unasuke:
To make bin/itamae executable.

before/after if here :arrow_down: 

Screenshot from 2019-05-16 17-38-19

### DIFF
--- a/bin/itamae
+++ b/bin/itamae
@@ -1,5 +1,7 @@
 #!/usr/bin/env ruby
 
+lib = File.expand_path('../lib', __dir__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+
 require 'itamae/cli'
 Itamae::CLI.start
-


bin/itamae
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env ruby
 
+lib = File.expand_path('../lib', __dir__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
```
sue445: @unasuke Hmm...

I think this is needless changes.

`bundler` doesn't add this line with `bundle gem --exe` 
https://github.com/bundler/bundler/blob/v2.0.1/lib/bundler/templates/newgem/exe/newgem.tt

So I don't want to add this.

If we run `bin/itamae` at local, add `bundle exec`

```bash
$ bundle exec ./bin/itamae --help
Commands:
  itamae destroy [cookbook|role] [NAME]   # Undo role or cookbook (short-cut alias: "d")
  itamae docker RECIPE [RECIPE...]        # Create Docker image
  itamae generate [cookbook|role] [NAME]  # Initialize role or cookbook (short-cut alias: "g")
  itamae help [COMMAND]                   # Describe available commands or one specific command
  itamae init NAME                        # Create a new project
  itamae jail RECIPE [RECIPE...]          # Run Itamae in jail
  itamae local RECIPE [RECIPE...]         # Run Itamae locally
  itamae ssh RECIPE [RECIPE...]           # Run Itamae via ssh
  itamae version                          # Print version
```

unasuke: Ok. I agree with add `bundle exec` to executing `bin/itamae` :smile:

PR #291, opened 2019-05-16
Ignore .ruby-version in this repository

unasuke:
In the case of gem development, we use several ruby versions that switch by rbenv.

So, `.ruby-version` files change but we want to ignore this from git.


unasuke:
Thx!

PR #296, opened 2019-07-27
Check http status code in http_request resource

takumin:
The current Itamae does not support HTTP status codes 4xx and 5xx.

This fix corresponds to HTTP status code 4xx and 5xx.


sue445:
@takumin Release as `v1.10.5`

takumin:
@sue445 @unasuke Thank you a lot!!!

PR #310, opened 2020-03-26
Improve `file` resource performance

labocho:
Now, `Itamae::Resource::File#pre_action` always transfer file content. But when the file is not modified, it is not necessary and spends time.

In this pull request, if the file exists on remote and `sha256sum` command is available, compare digest and avoid unnecessary file transfer.

30%+ faster in the benchmark below.
https://gist.github.com/labocho/23f8e089205a507e6089ce1e8ce5a863

### DIFF
--- a/lib/itamae/resource/file.rb
+++ b/lib/itamae/resource/file.rb
@@ -9,6 +9,10 @@ module Itamae
       define_attribute :group, type: String
       define_attribute :block, type: Proc, default: proc {}
 
+      class << self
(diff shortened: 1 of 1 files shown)


unasuke:
@labocho Could you squash those commits?

labocho:
@unasuke Yes! Done.

sue445:
I released as `v1.10.7` 💎

labocho:
Thanks!

PR #313, opened 2020-04-29
Simplify Git resource's get_revision method

pocke:
This pull request simplifies `get_revision` method of `Itamae::Resource::Git`.
It will improve performance.

This pull request contains two changes.

First, it replaces `git rev-list` with `git rev-parse`.
`git rev-parse` is faster, and enough. `rev-list` prints all revisions in the specified branch, but `rev-parse` only prints the head revision.

It is a really simple benchmark.

```ruby
require 'benchmark'
Benchmark.bm do |x|
  x.report{100.times{`git rev-parse @`}}
  x.report{100.times{`git rev-list @`}}
end

# In itamae repository
#
#        user     system      total        real
#    0.012472   0.020599   0.245997 (  0.235425)
#    0.124308   0.398403   2.054095 (  1.553693)
```

Second, it reduces git command executions when `git fetch` is unnecessary.

---

I guess this pull request only has a small impact on performance because `git` resource always try fetching new commits from the remote repository by default.

### DIFF
--- a/lib/itamae/resource/git.rb
+++ b/lib/itamae/resource/git.rb
@@ -88,7 +88,7 @@ module Itamae
         return result.stdout.strip if result.exit_status == 0
 
         fetch_origin!
-        run_command_in_repo("git rev-parse #{shell_escape(branch)}").strip
+        run_command_in_repo("git rev-parse #{shell_escape(branch)}").stdout.strip
       end
 
       def fetch_origin!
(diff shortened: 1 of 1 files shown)


pocke:
I unexpectedly removed `stdout` method call, so I've fixed it in cb116c2031c98da7e362645d46814ae4cc5e96be. :see_no_evil: 
I'll squash the commits if you need it.

unasuke:
@pocke Thank you! I'll release a new version after merge your pull requests.

PR #317, opened 2020-06-24
Make output unbuffered

terceiro:
This makes it possible to pipe the output of itamae to other tool and
have the output as it happens. Otherwise, the output is buffered and is
only available after itamae finishes.

### DIFF
--- a/lib/itamae/logger.rb
+++ b/lib/itamae/logger.rb
@@ -126,6 +126,7 @@ module Itamae
   @logger = ::Logger.new($stdout).tap do |l|
     l.formatter = Itamae::Logger::Formatter.new
   end.extend(Itamae::Logger::Helper)
+  $stdout.sync = true
 
   class << self
     def logger


unasuke:
@terceiro Thanks! Could tell me concretely usage of the "pipe the output of itamae to other tool"?

Sync stdout makes a little slow but it's not such a big issue I though. I want to another opinion.

terceiro:
> @terceiro Thanks! Could you tell me concretely usage of the "pipe the output of itamae to other tool"?

`itamae [...] | cat` is enough to reproduce this issue

> Sync stdout makes a little slow but it's not such a big issue I though. I want to another opinion.

well it's already the default when stdout is a tty, and also the time it takes to actually process each resource in a recipe is probably orders or magnitude larger than the time it takes to flush the corresponding logs to a non-tty stdout.

terceiro:
the travis failure (Itamae::Resource::HttpRequest::HTTPClientError) seems unrelated to my change. is that the case, or am I missing something?

unasuke:
> `itamae [...] | cat` is enough to reproduce this issue

Yes, I know how to "get" itamae's output from stdout, I want to know "what" to use itamae's output.

 The travis failure is httpbin.org's issue. :wink:

unasuke:
It's just my curiosity.

terceiro:
> Yes, I know how to "get" itamae's output from stdout, I want to know "what" to use itamae's output.

ah, ok. sorry.

I'm the author of a tool called [chake](https://gitlab.com/terceiro/chake) which I created originally to provide a way of using Chef™ without having a Chef™ server; I am now migrating my stuff to itamae, and adding support for it in chake. chake supports applying config to multiple servers in  parallel, and for that it runs the underlying tool with its output to a pipe, so that it can display the output prefixed with the hostname while the tool does its job on each server.

sue445:
I released as `v1.10.10`

PR #325, opened 2020-12-14
file: add support for sensitive files

terceiro:
sensitive files will not get changes to their contents logged.

### DIFF
--- a/lib/itamae/resource/file.rb
+++ b/lib/itamae/resource/file.rb
@@ -4,6 +4,7 @@ module Itamae
       define_attribute :action, default: :create
       define_attribute :path, type: String, default_name: true
       define_attribute :content, type: String, default: nil
+      define_attribute :sensitive, default: false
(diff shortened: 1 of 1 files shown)


unasuke:
@terceiro That's a good idea. In my opinion, print "diff existing but not show content because it's sensitive" in `show_content_diff` is more friendly. What do you think about it? 👀 

https://github.com/itamae-kitchen/itamae/blob/0d60ce7baa411f12ba2bb16c5f39dbd0a29d3566/lib/itamae/resource/file.rb#L158-L179

terceiro:
@unasuke yes I agree, that is better. Modified the PR accordingly.

unasuke (approved):
@terceiro Sorry, late. I checked changes and behavior on my local machine. It's looking good to me.

unasuke:
Released as v1.11.0

PR #332, opened 2021-02-07
Allow defining top-level modules without ::

### LINKED ISSUE #308: Capturing local variable `context` doesn't work
```rb
define :foo do
  context = params[:name]
  local_ruby_block 'foo' do
    block do
      p context
    end
  end
end

foo 'foo'
foo 'bar'
```

```
$ RBENV_VERSION=2.6.4 itamae local a.rb
 INFO : Starting Itamae...
 INFO : Recipe: /home/k0kubun/tmp/a.rb
"bar"
"bar"
```

The first `"bar"` should be `"foo"`, shouldn't it?

k0kubun:
This is a feature originally proposed in https://github.com/itamae-kitchen/mitamae/issues/64, using the idea in https://dev.to/okuramasafumi/eval-ing-twice-to-define-toplevel-module-from-string-33ma. 

Instead of `module ::Helper`, you'll be able to write the following recipes without errors:

```ruby
# helper.rb
module Helper
  def self.help
    :help
  end
end
```

```ruby
# recipe.rb
include_recipe 'helper'
Helper.help
```

This also fixes #308.


unasuke:
It seems ok. Ruby is very profound...Ah, where is the result of travis-ci.com?
Last night, we reached the quota of travis ci maybe. sue445-san had contacted travis. I want to wait for the response.

sue445:
@k0kubun I revived CI and merged the latest master.

But the build seems to be failing.
Please fix this 🙏

sue445 (approved):
LGTM

k0kubun:
Thanks for preparing GitHub Actions! :clap: :clap: :clap:

sue445:
released as `v1.12.1` 🚀

PR #336, opened 2021-02-17
Add weekly build

sue445:


.github/workflows/test.yml
```diff
@@ -9,6 +9,8 @@ on:
       - opened
       - synchronize
       - reopened
+  schedule:
+    - cron: "0 0 * * 5" # JST 9:00 (Sat)
```
sue445, line 13: **[Consultation]** This setting is sloppy :tired_face:
Do you have any good idea?

unasuke: What is this "sloppy"? It looks like no problem... Could you explain how is this "sloppy"? 👀

sue445: I want to run a scheduled build once a week.
But I can't think of a good schedule. (e.g. week, time)

unasuke: Hmm... ☺️

unasuke (approved):
🙆

PR #337, opened 2021-03-14
package: add :upgrade action

terceiro:
Since there is no generic way of checking whether there is a new version
of a given package, the upgrade action just installs the package
unconditionally, assuming that the package installation command will
install a newer version if it's available (that's the case for e.g.
apt-get).

Since it's only possible to check whether there was any version change
_after_ the action, show_differences needs to be called explicitly when
the package version changes for us to get some feedback in the logs.

### DIFF
--- a/lib/itamae/resource/package.rb
+++ b/lib/itamae/resource/package.rb
@@ -32,6 +32,16 @@ module Itamae
         end
       end
 
+      def action_upgrade(action_options)
+        run_specinfra(:install_package, attributes.name, attributes.version, attributes.options)
+        attributes.version = run_specinfra(:get_package_version, attributes.name).stdout.strip
+        attributes.installed = run_specinfra(:check_package_is_installed, attributes.name)
+        if !current.installed || attributes.version != current.version
+          show_differences
+          updated!
+        end
+      end
+
       def action_remove(action_options)
         if current.installed
           run_specinfra(:remove_package, attributes.name, attributes.options)


unasuke:
First, thank you for your pull request.

I'm still not sure why this is necessary.
A "configuration management tool" like Itamae needs to be idempotent, reproducible, and declarative.
In this respect, ":upgrade action" seems very procedural to me.

For example, If you need to install newer package, you can specify the "version" attribute to the package resource I think.

Can you give us a more detailed use case?

terceiro:
I maintain a platform of 30+ machines, which I'm migrating from chef to itamae. There are 3 packages that are developed for that platform, of which we always want the latest version. Today, those packages are uploaded to repositories that are configured in the machines, and in `chef`, the `install` action does the upgrade, and testing the migration I noticed that itamae skips the installation completely if _any_ version of the package is already installed.

terceiro:
Also, if I add new machines to the platform, using `install` the new ones will get the new version of those packages, while the older ones would be stuck with the previous version.

terceiro:
On the other hand, I understand your reluctance since my implementation violates a few principles of the itamae design, such as knowing what would happen without doing anything (e.g. --dry-run)

unasuke:
I understood. In that case, it's useful. And hard to write a spec. (I can't find a way to mock the behavior of the package manager.)
Could you give us more time to review it?

I want your opinion @sue445 . I basically agree with this.

sue445:
> I basically agree with this.

me too.

@terceiro Please add test code for this new feature.

PR #339, opened 2021-11-18
Fix Chef install script URL

NPoi:
This is a change to fix a problem where the ohai option was not working because the URL for the Chef installation had changed.


unasuke:
I confirmed it here https://docs.chef.io/chef_install_script/#unix-linux-and-macos Thanks!

image

PR #347, opened 2022-04-17
Support Psych4.x

chaaaaarlotte:
Since Psych4.0, YAML.load defaults to YAML.safe_load.

FYI https://github.com/ruby/psych/pull/487


unasuke (approved):
Thankyou!

sue445 (approved):

PR #292, opened 2019-05-18
Specify required ruby version to 2.2.6 or greater

unasuke:
Itamae is dependent by specinfra that's version is 2.64.0 or greater currently.
Specinfra requires net-ssh gem version 2.7 or greater and net-ssh gem requires ruby version 2.2.6 or greater.
Therefore itamae requires ruby 2.2.6 or greater as practically.
This pull request does specify the required ruby version to gemspec for end-user helpful.

## supplement
My investigation article (ja_JP) :arrow_right: https://note.mu/unasuke/n/n5fe89607c517

PR #351, opened 2022-05-22
Support repeatable "node_json" and "node_yaml" options

ineu:
Thor supports two ways to pass mutiple values to a flag: via the :array type or via the "repeatable" option. This PR adds the latter.
It allows to pass multiple `node_yaml` and `node_json` files that will all be read and applied to a node.

The logic is as follows:
- Hashes are merged
- Scalars and arrays are overwritten

A common use case for this feature is applying vars to a group of nodes (compare to Ansible's `group_vars` and `host_vars`).

Example:
```sh
  itamae ssh --node-json vars/global.json --node-yaml group.yml --node-yaml node-1.yml recipes.rb
```

### DIFF
--- a/itamae.gemspec
+++ b/itamae.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ["lib"]
 
-  spec.add_runtime_dependency "thor"
+  spec.add_runtime_dependency "thor", ">= 1.0.0"
   spec.add_runtime_dependency "specinfra", [">= 2.64.0", "< 3.0.0"]
   spec.add_runtime_dependency "hashie"
   spec.add_runtime_dependency "ansi"
--- a/lib/itamae/cli.rb
+++ b/lib/itamae/cli.rb
@@ -14,8 +14,8 @@ module Itamae
 
     def self.define_exec_options
       option :recipe_graph, type: :string, desc: "[EXPERIMENTAL] Write recipe dependency graph in DOT", banner: "PATH"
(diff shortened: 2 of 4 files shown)


unasuke (approved):
I think it's good. After merging this, we should increment minor version, don't we? @sue445

sue445 (changes requested):
@ineu `repeatable` option is available since thor v1.0.0+.
https://github.com/rails/thor/releases/tag/v1.0.0

Could you add `>= 1.0.0` to https://github.com/itamae-kitchen/itamae/blob/v1.12.6/itamae.gemspec#L29 ?

e.g. 

```ruby
spec.add_runtime_dependency "thor", ">= 1.0.0"
```

Others changes are LGTM

ineu:
Good point, I'll fix it.

Btw, should Ruby 2.3 still be supported? It has been deprecated for 6 years by now: https://endoflife.date/ruby
It's the default version in Ubuntu 16.04 LTS, but 18.04 LTS has 2.5

unasuke:
@ineu It's a sharp and interesting question for us.
Usually, no one uses ruby 2.3 now. But in the world of infrastructure as code, it's not surprising to use outdated instances like running ruby 2.3. Actually, specinfra which is the backend library of itamae, supports ruby 2.1.
And, Itamae has not specificated its support versions of ruby before now. If dropping support of eol-ed ruby, we should release a new major version.

related #292

sue445 (approved):
LGTM!

sue445:
I published v1.13.0 🚀 

* https://github.com/itamae-kitchen/itamae/blob/master/CHANGELOG.md#v1130
* https://rubygems.org/gems/itamae/versions/1.13.0

PR #352, opened 2022-06-12
Use execute

kaiquekandykoga:
Use execute instead of excute


sue445:
@kaiquekandykoga thx!

PR #355, opened 2022-07-03
Execute local_ruby_block code inside of chdir-ed block if cwd presented

### LINKED ISSUE #353: `cwd` does not work in `local_ruby_block`
It seems that `cwd` does not work in `local_ruby_block`.

Minimal reproduce code:
```ruby
# /tmp/foo/sample.rb
local_ruby_block 'foo' do
  cwd '/tmp'
  block do
    system('pwd')
  end
end
```

```bash
$ itamae version
Itamae v1.13.0
$ pwd
/home
$ itamae local /tmp/foo/sample.rb
 INFO : Starting Itamae... 
 INFO : Recipe: /tmp/foo/sample.rb
/home  # Expected: /tmp
```

The wiki says that `cwd` can be specified.
https://github.com/itamae-kitchen/itamae/wiki/local_ruby_block-resource

Is this a bug or did I use it incorrectly?

unasuke:
"local_ruby_block" execute by itamae's ruby process. The process doesn't change its working directory, which is presented by "cwd" attribute.
I think it's the root cause of issue #353.

I made the local_ruby_block code execute inside of Dir.chdir block if cwd attribute was presented to fix it.
This behavior has existed since 2015. (#146) There might be a recipe that is dependent on current behavior somewhere. Although I couldn't find public itamae recipes that use "local_ruby_block" with "cwd", the change has a small impact, I think.

Fix #353

### DIFF
--- a/lib/itamae/resource/local_ruby_block.rb
+++ b/lib/itamae/resource/local_ruby_block.rb
@@ -5,7 +5,13 @@ module Itamae
       define_attribute :block, type: Proc
 
       def action_run(options)
-        attributes.block.call
+        if attributes[:cwd]
+          Dir.chdir(attributes[:cwd]) do
+            attributes.block.call
+          end
+        else
+          attributes.block.call
+        end
       end
     end
   end
--- a/spec/integration/recipes/default.rb
+++ b/spec/integration/recipes/default.rb
@@ -527,6 +527,15 @@ local_ruby_block 'execute run_command' do
   end
 end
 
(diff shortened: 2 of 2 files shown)


sue445 (approved):
LGTM

unasuke:
Thanks. I will release this as v1.14.0 in a few days.

unasuke:
I will release itamae v1.14.0 (bump minor version) about 7/6 20:00 JST.

PR #363, opened 2023-06-03
require relative

kaiquekandykoga:
require cli using relative path making it possible to do changes in itamae and see the modifications running bin/itamae


unasuke:
@kaiquekandykoga I don't mean to comment intending to reject this change, but I think "bundle exec bin/itamae" would also work.
(We need to solve the problem of the test failing 😭  )

unasuke (approved):
That makes sense.

kaiquekandykoga:
Is it okay to ignore Ruby 2.7 --jit?
I tried to fix the CI using different approaches, and I am not sure what is the problem.

unasuke:
@kaiquekandykoga  Oops, my bad I forgot to mention that 💦  It appears that CI is not down due to your changes. So I would appreciate it if you do not include any changes regarding CI.

kaiquekandykoga:
I can revert the modifications I made in CI.
I just do not quite understand how the broken CI is not related to this patch. Looking the last commit cbe6cd5823dc92e9e6b50a5ab38c81720fb0aa78 at master seems to be passing.

unasuke:
> I just do not quite understand how the broken CI is not related to this patch

Me too. The failing spec is about sending http request to external resource. I doubt a cause of failure is around it maybe, but I didn't confirm it yet.
Anyway, your changeset is not a cause of failing tests.

unasuke (commented):
I thought the change was only in bin/itamae. The scope of the change is larger than I had imagined.

@sue445 What do you think?

@kaiquekandykoga Could you squash commits?

sue445:
@unasuke I agree with you. I think it would be better to use a different PR to correct the CI.

kaiquekandykoga:
I made it bigger, because I noticed that using require_relative for files related to itamae would facilitate the debugging. In this way I can run bin/itamae to check code changes without having to install it.

ryotarai:
To test it locally, I recommend to use bundle exec. IMHO it is not a good idea to replace require with require_relative.

lib/itamae/resource.rb
```diff
@@ -41,7 +41,7 @@ def get_resource_class(method)
 
       def autoload_plugin_resource(method)
         begin
-          require "itamae/plugin/resource/#{method}"
+          require_relative "plugin/resource/#{method}"
```
unasuke, line 44: I'm worried about this change. It may break the behavior of loading the external gem plugin, right?

kaiquekandykoga:
Just trying to clarify.

master branch.

```
[kandy@freebsd ~/a/itamae-test]$ cat recipe.rb
directory "./foo_bar" do
  action :create
end
[kandy@freebsd ~/a/itamae-test]$ ../itamae/bin/itamae local recipe.rb
 INFO : Starting Itamae...
 INFO : Recipe: /usr/home/kandy/a/itamae-test/recipe.rb
 INFO :   directory[./foo_bar] exist will change from 'false' to 'true'
```

current branch.

```
[kandy@freebsd ~/a/itamae-test]$ cat recipe.rb
directory "./foo_bar" do
  action :create
end
[kandy@freebsd ~/a/itamae-test]$ ../itamae/bin/itamae local recipe.rb
"Kandy message"
 INFO : Starting Itamae...
 INFO : Recipe: /usr/home/kandy/a/itamae-test/recipe.rb
 INFO :   directory[./foo_bar] exist will change from 'false' to 'true'
```

I can insert messages to help me debugging. If this is not the preferred approach, I am happy in closing this patch. Thank you.

unasuke:
@kaiquekandykoga You can use `bundle exec ../itamae/bin/itamae local recipe.rb` without this "require_relative ..." change; that's what is saying my previous comment https://github.com/itamae-kitchen/itamae/pull/363#issuecomment-1574908215 

Do you have any reason for cannot use `bundle exec`?

kaiquekandykoga:
Even using bundle exec does not behave the same way. With require, it will load the installed gem. With require_relative, I have the flexibility to choose from where in this case the project itself. An easy way to check this is uninstalling the itamae and try to run bin/itamae. Only with current branch will work.

unasuke:
> Even using bundle exec does not behave the same way.

Really? I tried this diff, and it looks work fine to me.

```ruby
# test_recipe.rb
directory "./foo_bar" do
  action :create
end
```

```diff
diff --git a/bin/itamae b/bin/itamae
index a5e2032..8ba2c3f 100755
--- a/bin/itamae
+++ b/bin/itamae
@@ -1,4 +1,5 @@
 #!/usr/bin/env ruby
 
 require 'itamae/cli'
+pp "==================hello"
 Itamae::CLI.start
diff --git a/lib/itamae/resource/directory.rb b/lib/itamae/resource/directory.rb
index 79259cd..349e24d 100644
--- a/lib/itamae/resource/directory.rb
+++ b/lib/itamae/resource/directory.rb
@@ -17,6 +17,7 @@ module Itamae
       end
 
       def show_differences
+        pp "==================hello from Itamae::Resource::Directory#show_differences"
         current.mode    = current.mode.rjust(4, '0') if current.mode
         attributes.mode = attributes.mode.rjust(4, '0') if attributes.mode
 
```


```
$ bundle exec bin/itamae local test_recipe.rb --dry-run
"==================hello"
 INFO : Starting Itamae... (dry-run)
 INFO : Recipe: /home/unasuke/src/github.com/itamae-kitchen/itamae/test_recipe.rb
"==================hello from Itamae::Resource::Directory#show_differences"
 INFO :   directory[./foo_bar] exist will change from 'false' to 'true'
```

Even, itamae gem is installed.

```
$ gem list --local | grep itamae
itamae (1.14.1)
```

kaiquekandykoga:
I am sorry. It seems that I need to run bundle exec from the itamae project.
When I was running from another directory it was not working.

```
[kandy@freebsd ~/a/itamae-test]$ bundle exec ../itamae/bin/itamae local ../itamae/recipe.rb
"bin/itamae"
 INFO : Starting Itamae...
 INFO : Recipe: /usr/home/kandy/a/itamae/recipe.rb
[kandy@freebsd ~/a/itamae-test]$ cd ../itamae
[kandy@freebsd ~/a/itamae]$ bundle exec bin/itamae local recipe.rb
"bin/itamae"
:run
 INFO : Starting Itamae...
 INFO : Recipe: /usr/home/kandy/a/itamae/recipe.rb
:show_differences
```

unasuke:
Ah...yes. bundler needs Gemfile in current working directory.

unasuke:
@kaiquekandykoga I hope your coding successfully 🙏

kaiquekandykoga:
Thank you for the patience! We have been talking all day long.

PR #378, opened 2025-05-05
[DO NOT MERGE]workaround for rubygems v3.6.7

unasuke:
## What
workaround for rubygems v3.6.7


sue445:
@unasuke How is the progress on this patch?

Since next ruby minor or major version will be released soon (probably 2025/12/2x), I hope we can merge it before then.

unasuke:
@sue445 ALL GREEN 💚

PR #434, opened 2026-04-08
Prepare for supply chain attacks

sue445:
Recently, supply chain attacks via GitHub Actions have become a serious issue 😭 

In this PR, we have introduced the following configurations to mitigate the risk of supply chain attacks:

1. Pin actions to a full-length commit SHA with https://github.com/suzuki-shunsuke/pinact
    * ref. https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions
2. Configure a cooldown period for Dependabot
    * ref. https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,5 @@ updates:
     directory: /
     schedule:
       interval: monthly
+    cooldown:
+      default-days: 7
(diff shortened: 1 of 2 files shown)


unasuke (approved):
👍

PR #442, opened 2026-07-24
Run http_request integration tests against a local go-httpbin server

takumin:
## Summary

The `http_request` integration tests currently talk to the public
`httpbin.org` service. That service is flaky and, under the CI matrix,
receives enough simultaneous requests that it returns 500s. As a workaround
the suite forced `skip_http_request_test` to `true` across most of the matrix.

This PR replaces the external dependency with a locally managed
[`mccutchen/go-httpbin`](https://github.com/mccutchen/go-httpbin) container,
so the tests no longer depend on an external service and the skip switch can be
removed entirely.

## Changes

- Add a small module that manages a local `go-httpbin` container for the
  integration tests. The image is pulled from GHCR to avoid Docker Hub's
  anonymous pull rate limits in CI.
- Start/stop the server around both the docker and local integration runs
  (Rakefile wiring, workflow steps, and the local spec runner). The target is
  resolved from `HTTPBIN_URL`, defaulting to the local server, and the
  container network is shared with the local backend's container.
- Adjust the specs for go-httpbin's response format: send an explicit
  `Content-Type` on form-encoded POST/PUT requests so the body is parsed, and
  loosen the matchers for go-httpbin's array-valued query/header echoes.
- Drop the `skip_http_request_test` CI throttle now that the tests no longer
  hammer a shared public service.
- Remove the now-unreachable `SKIP_HTTP_REQUEST_TEST` guards from the recipes
  and the specs, and stop propagating the variable through `docker run` /
  `docker exec`.

## Notes for reviewers

Most of the recipe diff is reindentation from dropping the `unless` wrapper.
Reviewing with `?w=1` (whitespace ignored) makes it much smaller.

### DIFF
--- a/spec/integration/recipes/default.rb
+++ b/spec/integration/recipes/default.rb
@@ -193,35 +193,39 @@ end
 ######
 
 unless ENV["SKIP_HTTP_REQUEST_TEST"] == "true"
+  httpbin_url = ENV.fetch("HTTPBIN_URL") { "http://127.0.0.1:#{ENV.fetch('HTTPBIN_HOST_PORT', '8080')}" }
+
   http_request "/tmp/http_request.html" do
-    url "https://httpbin.org/get?from=itamae"
+    url "#{httpbin_url}/get?from=itamae"
   end
 
   http_request "/tmp/http_request_delete.html" do
     action :delete
-    url "https://httpbin.org/delete?from=itamae"
+    url "#{httpbin_url}/delete?from=itamae"
   end
 
   http_request "/tmp/http_request_post.html" do
     action :post
+    headers "Content-Type" => "application/x-www-form-urlencoded"
     message "love=sushi"
-    url "https://httpbin.org/post?from=itamae"
+    url "#{httpbin_url}/post?from=itamae"
   end
 
   http_request "/tmp/http_request_put.html" do
     action :put
+    headers "Content-Type" => "application/x-www-form-urlencoded"
     message "love=sushi"
-    url "https://httpbin.org/put?from=itamae"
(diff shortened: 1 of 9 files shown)


spec/integration/recipes/default.rb
```diff
@@ -193,35 +193,39 @@
 ######
 
 unless ENV["SKIP_HTTP_REQUEST_TEST"] == "true"
```
sue445: I think this can be deleted as well just like `skip_http_request_test`.

spec/integration/recipes/ordinary_user.rb
```diff
@@ -84,28 +84,30 @@
 ###
 
 unless ENV["SKIP_HTTP_REQUEST_TEST"] == "true"
```
sue445: Same to https://github.com/itamae-kitchen/itamae/pull/442/changes#r3650373160

spec/integration/recipes/default.rb
takumin: Good catch, thanks. Removed in 3cf89f2 — the recipes and the specs no longer reference `SKIP_HTTP_REQUEST_TEST` at all, and 02d3df2 also stops propagating the variable through `docker run` / `docker exec`.

Most of the resulting recipe diff is just reindentation from dropping the `unless` wrapper, so `?w=1` makes it much easier to read.

spec/integration/recipes/ordinary_user.rb
takumin: Removed here as well in 3cf89f2.

sue445 (approved):
thx!

PR #444, opened 2026-07-26
Pull Docker Hub images through a mirror in CI

takumin:
The integration jobs let `docker run` pull the test images implicitly from Docker Hub. When Docker Hub is unreachable, the job fails before any test runs:

    docker: Error response from daemon: Get "https://registry-1.docker.io/v2/": context deadline exceeded

This is a connection timeout, not a rate limit, and it happens intermittently. Pulls from ghcr.io in the same jobs succeed, so the problem is specific to Docker Hub.

This PR configures a Docker Hub mirror in GitHub Actions. The `integration-docker` and `integration-local` jobs copy the checked-in `.github/workflows/docker-daemon.json` over `/etc/docker/daemon.json` and restart the Docker daemon, following:
https://docs.cloud.google.com/artifact-registry/docs/pull-cached-dockerhub-images

Notes:

- The daemon pulls docker.io images from the mirror first. If the mirror cannot serve an image, it falls back to Docker Hub automatically.
- The setting applies only to docker.io images, so the go-httpbin image from ghcr.io is not affected.
- It also works for implicit pulls by `docker run`, so no test helper changes are needed.

### DIFF
--- a/.github/workflows/test_main.yml
+++ b/.github/workflows/test_main.yml
@@ -58,6 +58,11 @@ jobs:
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
+      - name: Pull Docker Hub images through mirror.gcr.io
+        run: |
+          sudo cp .github/workflows/docker-daemon.json /etc/docker/daemon.json
+          sudo systemctl restart docker
+
       - uses: ruby/setup-ruby@9eb537ca036ebaed86729dcb9309076e4c5c3b74 # v1.314.0
         with:
           ruby-version: ${{ inputs.ruby }}
@@ -111,6 +116,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Pull Docker Hub images through mirror.gcr.io
+        run: |
+          sudo cp .github/workflows/docker-daemon.json /etc/docker/daemon.json
+          sudo systemctl restart docker
 
       - uses: ruby/setup-ruby@9eb537ca036ebaed86729dcb9309076e4c5c3b74 # v1.314.0
         with:
--- a/.github/workflows/docker-daemon.json
+++ b/.github/workflows/docker-daemon.json
@@ -1,0 +1,3 @@
+{
+  "registry-mirrors": ["https://mirror.gcr.io"]
+}


sue445:
@takumin Thank you for patch

I'm a bit hesitant about addressing a middleware issue on the test helper side.
If I need to use a mirror when running `docker pull` , it might be simpler to just set it up in GitHub Actions like this:
https://docs.cloud.google.com/artifact-registry/docs/pull-cached-dockerhub-images

takumin:
Thanks for the review! I reworked the PR as suggested: the mirror is now configured in GitHub Actions instead of the test helpers. If the mirror cannot serve an image, the Docker daemon falls back to Docker Hub automatically.

takumin:
https://github.com/itamae-kitchen/itamae/actions/runs/30278512286/job/90018611980?pr=444

The failing job (`test_main (3.3) / integration-docker (ubuntu:trusty)`) looks
like an unrelated flake rather than a regression from this change.

In the `provision` step, the `execute[apt-get update]` exec never returned and
the docker-api call hit its read timeout:

    DEBUG : Executing `apt-get update`...
      [:post, "/exec/69cd6f00.../start", ...]        # 15:10:03
    docker-api-2.4.0/lib/docker/connection.rb:93:
      read timeout reached (Docker::Error::TimeoutError)   # 15:12:26

The exec produced no output before the timeout, so the log does not show what
stalled inside the container. For comparison, the same command finished in 24s
with exit 0 in the Ruby 3.2 job of the same run.

Why I don't think the mirror caused it:

- Every other `integration-docker (ubuntu:trusty)` job in the same run applied
  the same `registry-mirrors` setting and passed; only this one failed.
- `registry-mirrors` only affects image pulls performed by the daemon, not
  traffic originating inside a running container.
- The Docker Hub failures this PR targets surface as
  `context deadline exceeded` / `Client.Timeout exceeded` on the pull itself,
  which is a different error from this one.

Re-running the failed job should be enough. Making `apt-get update` in the
integration recipe more resilient could be a separate follow-up.

sue445:
> https://github.com/itamae-kitchen/itamae/actions/runs/30278512286/job/90018611980?pr=444
>
> The failing job (test_main (3.3) / integration-docker (ubuntu:trusty)) looks
like an unrelated flake rather than a regression from this change.

I re-ran the job, and the tests passed.

https://github.com/itamae-kitchen/itamae/actions/runs/30278512286?pr=444

.github/workflows/test_main.yml
```diff
@@ -58,6 +58,13 @@ jobs:
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
+      - name: Pull Docker Hub images through mirror.gcr.io
+        run: |
+          [ -s /etc/docker/daemon.json ] || echo '{}' | sudo tee /etc/docker/daemon.json > /dev/null
+          sudo jq '. + {"registry-mirrors": ["https://mirror.gcr.io"]}' /etc/docker/daemon.json | sudo tee /etc/docker/daemon.json.new > /dev/null
```
sue445: I think it would be simpler to just copy the JSON file from the repository rather than manually generating it inside the shell script.

e.g.

`.github/workflows/docker-daemon.json`

```json
{
  "registry-mirrors": ["https://mirror.gcr.io"]
}
```

```bash
sudo cp .github/workflows/docker-daemon.json /etc/docker/daemon.json
```

.github/workflows/test_main.yml
```diff
@@ -112,6 +119,13 @@ jobs:
...
+          sudo mv /etc/docker/daemon.json.new /etc/docker/daemon.json
```
sue445: Same to https://github.com/itamae-kitchen/itamae/pull/444#discussion_r3662351454

.github/workflows/test_main.yml
takumin: Thanks, done in f0b15ba. The daemon config is now a checked-in file and the step just copies it:

```yaml
- name: Pull Docker Hub images through mirror.gcr.io
  run: |
    sudo cp .github/workflows/docker-daemon.json /etc/docker/daemon.json
    sudo systemctl restart docker
```

The GitHub-hosted runner images do not ship `/etc/docker/daemon.json` (nothing in `actions/runner-images` creates it), so replacing the file outright is equivalent to the previous `jq` merge.

.github/workflows/test_main.yml
takumin: Done as well in f0b15ba — both jobs now copy the same file.

takumin:
I also squashed the branch into a single commit and dropped the test-helper changes (`tasks/docker_image.rb` and the `Rakefile` / `tasks/integration_local_spec.rb` edits) that were still left in the diff by mistake after the earlier rework. The PR now only touches `.github/workflows/`.

## The current CI failures are not caused by this PR

All 21 `integration-docker (ubuntu:trusty)` jobs fail, while `integration-local` (17/17) and `unit` (21/21) pass. Every failure is a `Docker::Error::TimeoutError` on an apt operation inside the trusty container.

The same failure reproduces on the previous commit of this branch. Run 30330094864 ran bcf00b9 — the exact commit that passed 21/21 in run 30278512286 yesterday — and it failed 21/21 as well:

| run | commit | time (UTC) | integration-docker |
| --- | --- | --- | --- |
| 30278512286 | bcf00b9 | 07-27 15:08 | 21/21 pass |
| 30330094864 | bcf00b9 | 07-28 04:55 | 21/21 fail |
| 30330399342 | f0b15ba | 07-28 05:01 | 21/21 fail |

apt on trusty has become very slow. Locally, on an unrelated network with no mirror configured at all:

    apt-get update        -> exit 0, 345s
    apt-get install ruby  -> exit 0,  69s

Both exceed the 60s docker-api read timeout, which is why the jobs die at `apt-get update` (or at `apt-get install ruby` on the older commit, where `apt-get update` happened to finish in 39s). For comparison, the same `apt-get update` took 24s on 07-25.

The mirror is not involved: `registry-mirrors` only affects image pulls performed by the daemon, not traffic from inside a running container, and `ubuntu:trusty` has identical manifest digests on docker.io, mirror.gcr.io and public.ecr.aws.

master should be affected the same way. I think re-running once the trusty archives recover is enough here, and making the trusty jobs resilient to slow apt would be a separate topic.

sue445:
I re-ran the job, and the tests passed.

https://github.com/itamae-kitchen/itamae/actions/runs/30330399342

sue445 (approved):
LGTM!

PR #445, opened 2026-07-28
Add --docker-exec-timeout option to the docker backend

takumin:
`itamae docker` aborts with `Docker::Error::TimeoutError` when a command produces no output for 60 seconds.

docker-api sets a read timeout on the exec API call only when a `wait` option is given (`Docker::Exec#start!`); otherwise Excon's 60 seconds default applies. That timeout bounds the gap between reads, not the total runtime of the command, so anything that stays quiet for a minute — `apt-get` against a slow mirror, a compilation, a large download — kills the run even though it would have succeeded.

`Specinfra::Backend::Docker#docker_run!` merges `docker_container_exec_options` into the `exec` call, which is where `wait` would come from, but itamae only ever passed `docker_container_create_options` down to specinfra, so the setting was unreachable.

This adds `--docker-exec-timeout SECONDS` to the `docker` subcommand and translates it into `docker_container_exec_options`. Leaving it unset keeps the current 60 seconds behaviour, so nothing changes for existing users.

Notes:

- The backend returns `nil` rather than an empty hash when the option is absent. specinfra resolves the setting as `@config[key] || Specinfra.configuration.send(key)`, and an empty hash is truthy enough to shadow that fallback.
- The integration tests pass `--docker-exec-timeout 600`, overridable through `ITAMAE_DOCKER_EXEC_TIMEOUT`. `integration-docker` also gained `timeout-minutes`, since a single command may now run far longer before it gives up.

Verification, against a container provisioned with `execute 'sleep 90'`:

| | result |
| --- | --- |
| without the option | `Docker::Error::TimeoutError` after 60s |
| `--docker-exec-timeout 120` | succeeds after 90s |

`rake spec:unit` and `rake spec:integration:docker` pass locally, and the full CI matrix is green on this branch.

This touches `.github/workflows/test_main.yml`, which #444 also touches. The two changes are in different places within the file and merge cleanly in either order.

### DIFF
--- a/.github/workflows/test_main.yml
+++ b/.github/workflows/test_main.yml
@@ -45,6 +45,9 @@ jobs:
   integration-docker:
     runs-on: ubuntu-latest
 
+    # NOTE: --docker-exec-timeout raises the per-command timeout, so bound the job instead
+    timeout-minutes: 30
+
     strategy:
       fail-fast: false
 
--- a/Rakefile
+++ b/Rakefile
@@ -63,6 +63,7 @@ namespace :spec do
           cmd << "--container" << container_name
           cmd << "--tag" << "itamae:latest"
           cmd << "--tmp-dir" << (ENV['ITAMAE_TMP_DIR'] || '/tmp/itamae_tmp')
+          cmd << "--docker-exec-timeout" << (ENV['ITAMAE_DOCKER_EXEC_TIMEOUT'] || '600')
           cmd += suite
 
           p cmd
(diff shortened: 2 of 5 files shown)


takumin:
Closing this in favor of fixing the default instead.

The 60 seconds here is Excon's read timeout, so a command that produces no output for 60 seconds fails even while it is still running normally. `apt-get`, `make` and large `remote_file` downloads all hit this. That is a defect in the docker backend rather than something users should have to work around with a flag.

The Rakefile change in this PR makes that clear: it passes 600 unconditionally, which means the current default is not usable in practice.

`--docker-exec-timeout` also reads like a limit on total execution time while it actually bounds the time without output, and the documented `(default: 60)` is Excon's default rather than a value itamae guarantees. That is not a good contract to commit to a CLI option.

I will open a smaller PR that passes a sufficiently large wait to `docker_container_exec_options` by default, with no new CLI option and no Rakefile changes.

